PR #1, opened 2021-01-08
Adding all base files for meta-smartrack

HatsyRei:
Features include:
- Bitbake default build configuration
 - RCU base service recipe
 - RCU hostname service recipe and files
 - vsftpd recipe and configuration
 - Custom defconfig file
 - smartrack-minimal-image recipe

Practically all the same files in meta-smartrack-old.


buildconf/export
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+. ./oe-init-build-env ${BUILDDIR}
+
+echo ""
+echo "Toradex targets are:"
+echo "    tdx-reference-minimal-image"
+echo "    tdx-reference-multimedia-image"
+echo "    other (unsupported) targets may be found in"
+echo "    meta-toradex-demos/recipes-images/images/"
```
karhin1990: We may want to hide it as we don't want user to build toradex image and deploy it into Smart Racks RCU

buildconf/export
```diff
@@ -0,0 +1,46 @@
...
+	if grep -q "meta-toradex-bsp-common" ../.repo/manifest.xml && \
+	   ! grep -q "meta-toradex-bsp-common.git.*upstream" ../.repo/manifest.xml || \
+	   grep -qE "(integration.xml|next.xml)" ../.repo/manifest.xml
+	then
+		{
+			echo "# This is needed when building on integration. With use-head-next you"
+			echo "# always get the newest kernel. Without use-head-next your build may fail."
+			echo "MACHINEOVERRIDES =. \"use-head-next:\""
+			echo ""
+		} > conf/local.conf.new
```
karhin1990: Since we don't support building "integration", is this piece of code still relevant?

buildconf/export
```diff
@@ -0,0 +1,46 @@
...
+
+		cat conf/local.conf >> conf/local.conf.new
+		mv conf/local.conf.new conf/local.conf
+	fi
+
+	echo ""
+	$ECHO -e "\033[1mA sample conf/local.conf file has been created"
+	$ECHO -e "Check and edit the file to adapt to your local needs\033[0m"
+	echo "The following likely need your attention:"
+	echo "DL_DIR"
```
karhin1990: We may need a clearer description here, or directly omit it if we don't want our client to worry about this.

We should also have comments that guide our client how to build an image, e.g. "To build an image, execute: 'bitbake <supported_image_target>'

buildconf/local.conf
```diff
@@ -0,0 +1,269 @@
...
+MACHINE ?= "apalis-imx8x"
+#MACHINE ?= "apalis-tk1"
+#
+#MACHINE ?= "colibri-imx6"
+#MACHINE ?= "colibri-imx6ull"
+#MACHINE ?= "colibri-imx7"
+#MACHINE ?= "colibri-imx7-emmc"
+#MACHINE ?= "colibri-imx8x"
+#
+#MACHINE ?= "verdin-imx8mm"
```
karhin1990: We should remove other Toradex machine except apalis-imx8x because we don't want customer try to build those machine and run on our board,

recipes-kernel/linux/files/defconfig
```diff
@@ -0,0 +1,882 @@
+CONFIG_SYSVIPC=y
```
karhin1990, line 1: Can you briefly explain how this file get generated and what changes have you made on top of the toradex kernel config?

Also, should we name it as smart_racks_defconfig?

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+rcuserialpath=/usr/bin/rcu-serial
```
karhin1990, line 8: Can you put a TODO comment here indicates that this script/utility hasn't ready at the moment?

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -0,0 +1,27 @@
...
+if [ -f ${rcuserialpath} ]; then
+    serial=$(tr -d '\0' <${rcuserialpath})
+fi
+
+#if serial number is still empty
+if [ -z "$serial" -a "$serial" != " "  ]; then
+    serial="EMPTYSERIAL"
+fi
+
+hostname="NI-RCU-"${serial}
```
karhin1990: I would name it as NI-SmartRack-xxx because customer may not know what is actually an "RCU"

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
```
karhin1990, line 2: Can you put a comment on top of this file to mention the usage/purpose of this file?

recipes-kernel/linux/files/defconfig
HatsyRei: So far we haven't made any changes to the default kernel config yet (apart from an extra comment at the end of the file, to identify that this defconfig was used instead of the default one). We can change it to a different name, but this would most likely require some additional changes to be specified in our bbappend recipe as well. The KBUILD_DEFCONFIG_mx8 attribute in [Toradex's kernel recipe](https://git.toradex.com/cgit/meta-toradex-nxp.git/tree/recipes-kernel/linux/linux-toradex_5.4-2.1.x.bb?h=dunfell-5.x.y) points to 'defconfig' by default, so we'll have to try to change that.

recipes-core/rcu-hostname/files/rcuhostname
tanchunhau: Seems like can just handled by **if , elseIf , else** (according to priority - RCU serial, SoM serial , EmptySerial):
Please refer to below :

```
#set serial to RCU serial number
if [ -f ${rcuserialpath} ]; then
    serial=$(tr -d '\0' <${rcuserialpath})

#set serial to Apalis SOM serial number
elif [ -f ${tdxserialpath} ]; then
     serial=$(tr -d '\0' <${tdxserialpath})

#set serial to "EMPTYSERIAL"
else
    serial="EMPTYSERIAL"
fi
```

tanchunhau (commented):
Note: 3x3 bit flags for a mode:
https://stackoverflow.com/questions/18415904/what-does-mode-t-0644-mean/18415935

recipes-core/rcu-hostname/files/rcuhostname
HatsyRei: Referring to Toradex's template for this script, it seems like there's a chance our serial file will exist but somehow empty or corrupted. If we proceed with the if elif else construct, I'm afraid we might not be able to account for that case.

conf/layer.conf
```diff
@@ -0,0 +1,13 @@
+# We have a conf and classes directory, add to BBPATH
+BBPATH .= ":${LAYERDIR}"
+
+# We have recipes-* directories, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
+            ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "meta-smartrack"
+BBFILE_PATTERN_meta-smartrack = "^${LAYERDIR}/"
+BBFILE_PRIORITY_meta-smartrack = "24"
```
karhin1990: Can you place a brief comment here on how to come out this magic number (24)? It will help for future references and maintenance.

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -0,0 +1,30 @@
...
+# This script is responsible for reading the RCU serial and setting the RCU hostname accordingly. Paired with avahi-daemon which enables mDNS, this provides an option for clients to communicate with the RCU just by knowing its serial (instead of IP address).
```
karhin1990: ... the RCU Serial **Number** and setting ...

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -0,0 +1,30 @@
...
+if [ -f ${rcuserialpath} ]; then
+    serial=$(tr -d '\0' <${rcuserialpath})
+fi
+
+#if serial number is still empty
+if [ -z "$serial" -a "$serial" != " "  ]; then
+    serial="EMPTYSERIAL"
+fi
+
+hostname="NI-SmartRack-"${serial}
```
karhin1990, line 26: Refer to Jonathan's comment on the feature spec, we need a way for user to set their custom hostname. 
I believe user will set the hostname via RCU Service, perhaps you need some placeholder code to read from a file (where RCU Service will write the user-defined hostname). If this file exist and the content is valid, we will use it as the hostname instead.

karhin1990 (changes requested):
Overall looks solid, just minor things to address then we are good to go.

conf/layer.conf
HatsyRei: Added a comment above the line. I also changed priority from 24 to 25. My original intention was to allow meta-smartrack have priority over other Toradex layers. Layer priority 24 (assumed by meta-toradex-demos) was the highest among Toradex layers, so to achieve the aim I went one value higher.

recipes-core/rcu-hostname/files/rcuhostname
HatsyRei: @karhin1990 I added a todo in this file, as well as a new task under the RCU image user story in AzDO.

karhin1990 (approved):

PR #2, opened 2021-01-20
Inclusion of custom dt overlay into RCU image

HatsyRei:
The apalis-imx8x_smartrack_overlay.dts enables i2c buses and muxes certain pins to GPIO. Using a customized device-tree-overlays_git.bb (which shadows the one provided in meta-toradex-nxp), this overlay is installed into the RCU image bootfs partition. The customized apalis-imx8x.conf file in meta-smartrack/conf/machine then instructs overlay.txt file in the bootfs partition to point to our overlay.

### DIFF
--- a/conf/machine/apalis-imx8x.conf
+++ b/conf/machine/apalis-imx8x.conf
@@ -1,0 +1,73 @@
+#@TYPE: Machine
+#@NAME: Toradex Apalis iMX8X
+#@DESCRIPTION: Toradex Apalis iMX8X powered by a i.MX 8X SoC
+#@MAINTAINER: Philippe Schenker <philippe.schenker@toradex.com>
+
+MACHINE_NAME = "Apalis-iMX8X"
+
+# for C0 silicon add mx8qxpc0 to the right of mx8qxp
+SILICON-OVERRIDES ?= "mx8qxpc0:"
+MACHINEOVERRIDES =. "mx8:mx8x:mx8qxp:${SILICON-OVERRIDES}"
+
+IMX_DEFAULT_BSP = "nxp"
+
+require conf/machine/include/imx-base.inc
+# if one wants cortexa35-crypto, set DEFAULTTUNE accordingly
+DEFAULTTUNE_mx8qxp = "aarch64"
+require conf/machine/include/tune-cortexa35.inc
+
+# Don't include kernels in standard images
+RDEPENDS_${KERNEL_PACKAGE_NAME}-base = ""
+
+LOADADDR = ""
+
+# we do not want to have getty running on tty1 as we run
+# auto-serial-console there
+USE_VT = "0"
+
+KERNEL_DEVICETREE = " \
+    freescale/imx8qxp-apalis-eval.dtb \
+"
+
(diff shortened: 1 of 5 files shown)


conf/machine/apalis-imx8x.conf
```diff
@@ -0,0 +1,73 @@
+#@TYPE: Machine
```
karhin1990, line 1: As per what we talked offline, it would be better to define a "machine" for smart-racks as we create our own carrier board.

conf/machine/apalis-imx8x.conf
```diff
@@ -0,0 +1,73 @@
...
+IMAGE_FSTYPES += "teziimg"
+
+# The imx-boot container takes care for the i.MX 8 offset, so the container
+# has to be flashed at offset 0 directly
+UBOOT_BINARY_TEZI_EMMC = "imx-boot"
+OFFSET_BOOTROM_PAYLOAD = "0"
+
+TEZI_EXTERNAL_KERNEL_DEVICETREE_BOOT = "apalis-imx8x_smartrack_overlay.dtbo"
+
+TORADEX_PRODUCT_IDS = "0046 0053 0054 2600"
```
karhin1990, line 73: This seems to used in the original .conf file to specify which dtb to use for each product. Do we need to retain that id<->dtb mapping?

recipes-kernel/linux/device-tree-overlays_git.bb
```diff
@@ -0,0 +1,20 @@
+SUMMARY = "SmartRack device tree overlays"
+DESCRIPTION = "SmartRack device tree overlays from within the layer."
+
+SRC_URI = "file://apalis-imx8x_smartrack_overlay.dts file://template_overlay.dts"
+
+inherit devicetree
+
+S = "${WORKDIR}"
+
+COMPATIBLE_MACHINE = ".*(mx[678]).*"
```
karhin1990, line 10: we should restrict it to only compatible with our smart-rack machine, because the overlay was customized for smart-rack carrier board

HatsyRei:
Closing PR for now, will make a new one once the custom machine is complete and the kernel custom defconfig part is fixed.

karhin1990 (commented):
Just notice you are closing the review when I about to post by comment.

conf/machine/apalis-imx8x.conf
HatsyRei: Yeah... I wasn't sure what it actually meant by product ID here. I initially thought it was the product serial number, but what you said made better sense. I'll try dropping it and see if it causes any issues.

Update: Dropping it did cause issues. In [image_type_tezi.bbclass](http://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/classes/image_type_tezi.bbclass?h=dunfell-5.x.y), having an absent product ID causes a bbfatal to be called. I also missed out 4 lines below the product id declaration (the id to dtb assignment) from the original .conf file, but no error was reported because that was used for to write dtb to raw nand. Since the Apalis iMX8X comes with eMMC, it did not report any issues.

PR #3, opened 2021-01-25
Merging latest changes into base

HatsyRei:
Changes include:
 - Using .cfg file for kernel customization instead of custom defconfig
 - Device tree overlay application to base RCU image
 - Renaming of "Smart Rack" to "Smart Racks" for all mentioned instances
 -  Machine customization from "apalis-imx8x" to "apalis-imx8x-smartracks"


conf/machine/apalis-imx8x-smartracks.conf
```diff
@@ -0,0 +1,73 @@
+#@TYPE: Machine
+#@NAME: Toradex Apalis iMX8X
+#@DESCRIPTION: Toradex Apalis iMX8X powered by a i.MX 8X SoC
+#@MAINTAINER: Philippe Schenker <philippe.schenker@toradex.com>
+
+MACHINE_NAME = "Apalis-iMX8X"
```
karhin1990: I thought we gonna define a new MACHINE_NAME?

recipes-kernel/linux/linux-toradex_5.%.bbappend
```diff
@@ -1,3 +1,3 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
-SRC_URI += "file://defconfig"
+SRC_URI += "file://smartracks.cfg"
```
karhin1990: Can you add a comment here stating this cfg will be applied on which defconfig? Mainly for educational/documentation purpose and ease future development

Example: "This cfg file will be applied on top of the default Apalis iMX8X defconfig, which specified in meta-toradex-nxp/recipes-kernel/linux/linux-toradex_<version>.bb"

conf/machine/apalis-imx8x-smartracks.conf
```diff
@@ -0,0 +1,73 @@
...
+IMAGE_FSTYPES += "teziimg"
+
+# The imx-boot container takes care for the i.MX 8 offset, so the container
+# has to be flashed at offset 0 directly
+UBOOT_BINARY_TEZI_EMMC = "imx-boot"
+OFFSET_BOOTROM_PAYLOAD = "0"
+
+TEZI_EXTERNAL_KERNEL_DEVICETREE_BOOT = "smartracks_overlay.dtbo"
+
+TORADEX_PRODUCT_IDS = "0046 0053 0054 2600"
```
karhin1990, line 73: Should we still keep this line? I can't really find from web about its usage though..

recipes-core/hello/hello.bb
```diff
@@ -7,8 +7,10 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "hello.service"
 
-SRCREV = "${AUTOREV}"
+# TODO: Alter SRC_URI location once rcu-service repo is moved to either AzDO or GitHub/NI
 SRC_URI = "git://github.com/NI-SmartRack/rcu-service.git;branch=main"
```
karhin1990: Since we have moved to github/ni organization, I think you can update it now

recipes-core/hello/hello.bb
```diff
@@ -7,8 +7,10 @@ inherit systemd
...
+SRCREV = "${AUTOREV}"
```
karhin1990: Can you add a TODO comment here to indicate that we will not be using AUTOREV once User Story 1240399 is completed.

recipes-core/hello/hello.bb
```diff
@@ -7,8 +7,10 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "hello.service"
```
karhin1990: We shall update the name of this file to rcu-service_<version>.bb. Also, line 1 and 2 of this file need to update as well

recipes-core/hello/hello.bb
```diff
@@ -7,8 +7,10 @@ inherit systemd
...
```
karhin1990: service name has to update

recipes-core/hello/hello.bb
```diff
@@ -7,8 +7,10 @@ inherit systemd
...
 
-SRCREV = "${AUTOREV}"
+# TODO: Alter SRC_URI location once rcu-service repo is moved to either AzDO or GitHub/NI
 SRC_URI = "git://github.com/NI-SmartRack/rcu-service.git;branch=main"
+SRCREV = "${AUTOREV}"
+
 FILES_${PN} += "${systemd_unitdir}/system/hello.service"
```
karhin1990: service name has to update

recipes-kernel/linux/linux-toradex_5.%.bbappend
```diff
@@ -1,3 +1,3 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
karhin1990: Any reason to use wildcard (%) in the filename? Shall we name this file as linux-toradex-5.4-2.1.x.bbappend instead?

conf/machine/apalis-imx8x-smartracks.conf
```diff
@@ -0,0 +1,73 @@
...
+RDEPENDS_${KERNEL_PACKAGE_NAME}-base = ""
+
+LOADADDR = ""
+
+# we do not want to have getty running on tty1 as we run
+# auto-serial-console there
+USE_VT = "0"
+
+KERNEL_DEVICETREE = " \
+    freescale/imx8qxp-apalis-eval.dtb \
```
karhin1990, line 29: Hmm.. it seems odd that smart-racks machine is using apalis evaluation board's device tree.. I learnt that we use overlay to *transform* apalis eval board to our carrier board, but not sure is it really the right practice. 
TMK overlay is mainly to alter (enable/disable) certain functionality of a specific board - meaning we will have a dedicated dtb for a specific board, and only use overlay if we want to tweak certain functionalities/small portion of the board. Thoughts?

recipes-kernel/linux/files/template_overlay.dts
```diff
@@ -0,0 +1,6 @@
+/dts-v1/;
```
karhin1990, line 1: what is the usage of this file?

recipes-kernel/linux/files/smartracks_overlay.dts
```diff
@@ -0,0 +1,173 @@
...
+            #address-cells = <1>;
+            #size-cells = <0>;
+            pinctrl-names = "default";
+            pinctrl-0 = <&pinctrl_lpi2c5>;
+            clock-frequency = <100000>;
+            status = "okay";
+        };
+    };
+
+    /* Enable interrupt parent for i2c_mipi_csi0 */
```
karhin1990, line 173: Would be good to have such comment on each fragments listed in this file, to ease the reader to understand what each fragment is trying to do.

karhin1990 (changes requested):

conf/machine/apalis-imx8x-smartracks.conf
HatsyRei: In [image_type_tezi.bbclass](http://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/classes/image_type_tezi.bbclass), around line 286, bb.fatal is called in case no product ID is found, so while we don't really use this to write any dtb to raw nand, the product ID still has to be specified.

conf/machine/apalis-imx8x-smartracks.conf
HatsyRei: Yep. Seems like I overlooked this one. Managed to change this in my previous build and it worked without issues.

recipes-kernel/linux/files/template_overlay.dts
HatsyRei: Originally added to facilitate adding new overlays in case we plan on writing any. My plan was to separate overlays used to enable i2c buses and other overlays (such as ones to alter ethernet, change SD/MMC pins) but seems like it may be less relevant now.

conf/machine/apalis-imx8x-smartracks.conf
karhin1990: Ah I see. Alright I will mark this comment as Resolved

recipes-kernel/linux/files/template_overlay.dts
karhin1990: Oh I see. I'm fine if you prefer to keep it, just make sure there is comment to describe the file usage then I'm good.

conf/machine/apalis-imx8x-smartracks.conf
HatsyRei: You are correct. By right, if we are creating a new board for the Apalis-iMX8X, we should create a new board level device tree. It would be structurally similar to the [eval board device tree](https://git.toradex.com/cgit/linux-toradex.git/tree/arch/arm64/boot/dts/freescale/imx8qxp-apalis-eval.dts?h=toradex_5.4-2.1.x-imx), which includes the [imx8qxp-apalis.dtsi](https://git.toradex.com/cgit/linux-toradex.git/tree/arch/arm64/boot/dts/freescale/imx8qxp-apalis.dtsi?h=toradex_5.4-2.1.x-imx) containing SOM level declarations. However, the issue is that some I2C and GPIO pin configurations we need to customize according to our custom board pinout design is declared in this SOM level device tree (under the iomuxc node) and not the board level device tree. For this use case,  most Toradex documentation suggests using overlays while leaving the board level device tree intact. 

I think we will eventually create our own board level device tree to replace the eval board device tree, though in the meantime we probably will continue to use the overlay for pinmuxing while we figure out how to do that.

karhin1990: Thanks for explaining, Yan Huan. In this case, can you help to create a "Technical Debt" work item under "Project Groundwork" feature in AzDO to track it? I agree with you that we can revisit this later.

recipes-core/rcu-service/rcu-service_.bb
```diff
@@ -0,0 +1,29 @@
...
+S = "${WORKDIR}/git"
+TARGET_CC_ARCH += "${LDFLAGS}"
+
+do_compile() {
+         ${CXX} rcu-service.cpp -o rcu-service
+}
+
+do_install() {
+         install -d ${D}${bindir}
+         install -m 0755 hello ${D}${bindir}
```
karhin1990: binary name not updated

recipes-core/rcu-service/rcu-service_.bb
```diff
@@ -0,0 +1,29 @@
+SUMMARY = "RCU Service"
```
karhin1990: come to think of it, seems like we need include the version in the recipe filename, else we will have no way to know what is version of "rcu-service" is currently installed into the image

karhin1990 (changes requested):
Last 2 comments then we are good to go~

recipes-core/rcu-service/rcu-service_.bb
HatsyRei: Added version number 1.0 to the recipe for now.

karhin1990 (approved):
Great work!

PR #5, opened 2021-02-08
Update rcu-service_1.0.bb

tanchunhau:
make it ready for build by using externalsrc  (in azure pipeline)

### DIFF
--- a/recipes-core/rcu-service/rcu-service_1.0.bb
+++ b/recipes-core/rcu-service/rcu-service_1.0.bb
@@ -9,8 +9,7 @@ SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
 SRC_URI = "git://github.com/ni/rcu-service.git;branch=main"
 
-# TODO: AUTOREV will no longer be used once User Story 1240399 is completed
-# https://dev.azure.com/ni/DevCentral/_workitems/edit/1240399/
+# TODO: Assign to a fixed revision after release
 SRCREV = "${AUTOREV}"
 
 FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
@@ -18,12 +17,12 @@ S = "${WORKDIR}/git"
 TARGET_CC_ARCH += "${LDFLAGS}"
 
 do_compile() {
-         ${CXX} rcu-service.cpp -o rcu-service
+         ${CXX} ${S}/rcu-service.cpp -o rcu-service
 }
 
 do_install() {
          install -d ${D}${bindir}
          install -m 0755 rcu-service ${D}${bindir}
          install -d ${D}/${systemd_unitdir}/system
-         install -m 0644 ${WORKDIR}/git/rcu-service.service ${D}/${systemd_unitdir}/system
+         install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
 }
--- a/recipes-core/rcu-service/rcu-service_1.0.bb
+++ b/recipes-core/rcu-service/rcu-service_1.0.bb
@@ -7,6 +7,8 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
+# TODO: Point to release branch when it is ready.
+# If EXTERNAL_SRC was specified, it will take precedence of SRC_URI and SRC_REV
 SRC_URI = "git://github.com/ni/rcu-service.git;branch=main"
 
 # TODO: Assign to a fixed revision after release


HatsyRei (approved):

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -9,21 +9,20 @@ SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
 SRC_URI = "git://github.com/ni/rcu-service.git;branch=main"
```
karhin1990, line 12: Can you help to add a TODO here saying: "TODO: Point to release branch when it is ready", and also a comment of "If EXTERNAL_SRC was specified, it will take precedence of SRC_URI and SRC_REV"

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -9,21 +9,20 @@ SYSTEMD_SERVICE_${PN} = "rcu-service.service"
...
+# TODO: Assign to a fixed revision after release
 SRCREV = "${AUTOREV}"
 
 FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
 S = "${WORKDIR}/git"
 TARGET_CC_ARCH += "${LDFLAGS}"
 
 do_compile() {
-         ${CXX} rcu-service.cpp -o rcu-service
+         ${CXX} ${S}/rcu-service.cpp -o rcu-service
```
karhin1990, line 22: Just curious, why previously it can compile without specifying ${s} ? Any reason we want to make this change?

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -9,21 +9,20 @@ SYSTEMD_SERVICE_${PN} = "rcu-service.service"
...
 }
 
 do_install() {
          install -d ${D}${bindir}
          install -m 0755 rcu-service ${D}${bindir}
          install -d ${D}/${systemd_unitdir}/system
-         install -m 0644 ${WORKDIR}/git/rcu-service.service ${D}/${systemd_unitdir}/system
+         install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
```
karhin1990, line 29: If we are using EXTERNAL_SRC, ${S} will be undefined (as your pipeline will comment it out).  Will it cause this line to fail?

tanchunhau: ${S} (source files) and ${B} (build artifacts) is already predefined. 
visit this link :
https://www.yoctoproject.org/docs/1.8/ref-manual/ref-manual.html
search for :
5.7.3. Kernel Build Changes

recipes-core/rcu-service/rcu-service_1.0.bb
tanchunhau: without specifying is relative directory, it will fail for externalsrc compilation.
Here I make it absolute path, since the ${S} is dynamically assign in local build & pipeline build. It pass for both cases.

recipes-core/rcu-service/rcu-service_1.0.bb
karhin1990: What would be the default value for ${S}? I didn't manage to find it from the section 5.7.3 you linked.

recipes-core/rcu-service/rcu-service_1.0.bb
karhin1990: Nice. Thanks for explaining!

recipes-core/rcu-service/rcu-service_1.0.bb
tanchunhau: source files directory

karhin1990: I means the value of it. When we commented out S = "${WORKDIR}/git", and define EXTERNAL_SRC in local.conf, what will ${S} interpreted to?

HatsyRei: From the documentation on externalsrc in section 6.33, seems like the S variable is set by EXTERNAL_SRC. If I interpret it correctly, it means when we define the external source in local.conf, ${S} will give the same thing as ${EXTERNAL_SRC}.

https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#ref-classes-externalsrc

I'm not too sure, but looking at "s_dir", the linkage should happen around line 137 and 195 in externalsrc.bbclass here:
https://github.com/openembedded/openembedded-core/blob/master/meta/classes/externalsrc.bbclass

tanchunhau: when using externalsrc, the {S} = to externalsrc directory.

when not using externalsrc: (it will at the temp place. (checkout at temporary place))

{workdir} = .../build/tmp/work/(aarch64-tdx-linux)/rcu-service/1.0-r0
{S} = .../build/tmp/work/(aarch64-tdx-linux)/rcu-service/1.0-r0/rcu-service-1.0

recipes-core/rcu-service/rcu-service_1.0.bb
tanchunhau: ok

recipes-core/rcu-service/rcu-service_1.0.bb
karhin1990: Cool. Thanks guys!

karhin1990 (approved):
Good work!

PR #6, opened 2021-02-09
Pushing minor changes

HatsyRei:
- Remove packages for development use only (dtc, python) from smartracks image recipe.
- Add default password to root account of base image
- Fix use of https protocol for rcu-service fetching


karhin1990 (approved):
Approved!

PR #7, opened 2021-02-25
Edit recipes to allow building RCU Service and SDK

HatsyRei:
Also bumped git hash in smartracks-manifest, so bbappend recipe for Toradex linux kernel is renamed.

https://github.com/ni/smartracks-manifest/commit/0f8f4fbf08e3983a66f43bc21a6c2fc0288cddbe


recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,13 +18,29 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
 S = "${WORKDIR}/git"
 TARGET_CC_ARCH += "${LDFLAGS}"
 
-do_compile() {
-         ${CXX} ${S}/rcu-service.cpp -o rcu-service
+DEPENDS = " protobuf-native grpc grpc-native "
```
karhin1990, line 21: Double check: No RDEPEND needed?

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,13 +18,29 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
...
+
+inherit cmake pkgconfig
```
karhin1990: Question: pkgconfig is for what purpose?

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,13 +18,29 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
...
+
+# A hacky way to get grpc dependencies to work in order to build RCU service
```
karhin1990: Nit: Comment improvement
Perhaps can update this line to: "Workaround for grpc cross-compilation. This issue was reported in: [put the URL of someone ask this question in web]"

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,13 +18,29 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
...
+# Remove grpc installation .cmake files to allow cmake to find grpc-native 
+# this allows cmake to find grpc_cpp_plugin
+# Then, manually copy five aarch64 .so files required to build RCU service to x86 sysroot
```
karhin1990: Nit: 
- Capital 'S' for RCU [S]ervice
- Can you put a comment why these .so has to copied to x86 sysroot?

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,13 +18,29 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
...
+         cp ${WORKDIR}/recipe-sysroot/usr/lib/libgrpc++.so.1.24.3 ${WORKDIR}/recipe-sysroot-native/usr/lib/
+         cp ${WORKDIR}/recipe-sysroot/usr/lib/libgrpc.so.6.0.0 ${WORKDIR}/recipe-sysroot-native/usr/lib/
+         cp ${WORKDIR}/recipe-sysroot/usr/lib/libgpr.so.6.0.0 ${WORKDIR}/recipe-sysroot-native/usr/lib/
+         cp ${WORKDIR}/recipe-sysroot/usr/lib/libaddress_sorting.so.6.0.0 ${WORKDIR}/recipe-sysroot-native/usr/lib/
 }
 
 do_install() {
          install -d ${D}${bindir}
          install -m 0755 rcu-service ${D}${bindir}
+         install -m 0755 rcu-service-cpp-test-client ${D}${bindir}
```
karhin1990: We don't want the test client to be installed in RCU Image by default, because it's not needed by customer. 
It's for internal used only, so we should export it to a separate package (e.g. smartracks-tests.ipk) that available on the package feed.

karhin1990 (changes requested):

recipes-core/rcu-service/rcu-service_1.0.bb
HatsyRei: I think this is one of the things I tried in order to get gRPC to work, and it is accidentally left in there. I'll run a clean build without pkgconfig and flash the image, make sure everything works, then I'll remove it in case it is not actually required.

recipes-core/rcu-service/rcu-service_1.0.bb
HatsyRei: Yep, just clean built and flashed into SOM, can confirm that it is working. Another observation is, even without RDEPENDS, installing the ipk by itself will still cause opkg to look for gRPC and other runtime dependencies. Seems like runtime dependencies are implicitly specified somewhere in the gRPC recipes... Do you think it would be a good idea to specify RDEPENDS anyway? It does not seem to affect anything for now.

karhin1990: I think we need to know if there is actually grpc-specific runtime libraries (you could try ld the rcu-service and see if it shows you any dependency that provided by grpc). If yes, then we need to find out which OE package provides it and add it to RPDEPENDS; else we are safe to leave it as it is.

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -18,8 +18,23 @@ FILES_${PN} += "${systemd_unitdir}/system/rcu-service.service"
...
+DEPENDS = " protobuf-native grpc grpc-native "
+
+inherit cmake
+
+# Workaround for gRPC cross-compilation. This issue was reported in: https://github.com/grpc/grpc/issues/17708
+# We remove gRPC installation .cmake files in target sysroot (aarch64) to allow cmake to locate grpc_cpp_plugin in sysroot-native (x86_64)
+# Now that cmake locates gPRC installation in sysroot-native, the linker looks for _GRPC_GRPCPP, _REFLECTION .so files in sysroot-native lib directory
+# However, compiling RCU Service for ARM-based target requires these .so files to be built for aarch64
+# Hence, we solve this by copying the five aarch64 .so files below from target sysroot to sysroot-native, replacing their x86 counterpart. 
+
```
karhin1990: Nit: Extra line

recipes-core/rcu-service/rcu-service_1.0.bb
```diff
@@ -28,3 +43,9 @@ do_install() {
          install -d ${D}/${systemd_unitdir}/system
          install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
 }
+
+EXTRA_OECMAKE = " \
```
karhin1990: Would be helpful to put some comment for this. Like what you previously written in the rcu-service CMakelist.

Question: Is it matters where to specify this? Do it has to be specify before "inherit cmake" so that it visible to cmake during do_compile/do_configure?

karhin1990 (commented):
Will there be another PR to package rcu-service-cpp-test-client into a separate ipk?
Will be approving this PR once the RDEPENDS comment is addressed.

recipes-core/rcu-service/rcu-service_1.0.bb
HatsyRei: > Will there be another PR to package rcu-service-cpp-test-client into a separate ipk?
> Will be approving this PR once the RDEPENDS comment is addressed.

Yep. We can write a new recipe to package rcu-service-cpp-test-client into an ipk. It would contain the same lines used to build gRPC dependent applications, with no installation of .service or dependency on systemd. I think I can add that really soon if that is all we need to do.

HatsyRei: > Question: Is it matters where to specify this? Do it has to be specify before "inherit cmake" so that it visible to cmake during do_compile/do_configure?

Doesn't seem like it matters. I managed to move EXTRA_OECMAKE above the inherit statement and get it to build just fine.

recipes-core/rcu-service/rcu-service_1.0.bb
HatsyRei: I managed to find this link, which says the following. Could be our case here:
https://wiki.yoctoproject.org/wiki/TipsAndTricks/Packaging_Prebuilt_Libraries

>If the precompiled binaries are not statically linked and have dependencies on other libraries, then by adding those libraries to DEPENDS the linking can be examined and the appropriate RDEPENDS automatically added.

karhin1990: Cool, I'll mark this comment as Resolved. Thanks!

recipes-core/rcu-service/rcu-service_1.0.bb
karhin1990: Great, I'm going to approve this PR and please use a new PR for the tests-specific package creation. Thanks!

karhin1990 (approved):

PR #8, opened 2021-03-03
Pushing overlay fix and recipe for RCU Service cpp tester

HatsyRei:


recipes-core/rcu-service-cpp-test-client/rcu-service-cpp-test-client_1.0.bb
```diff
@@ -0,0 +1,46 @@
+SUMMARY = "RCU Service CPP Test Client"
```
karhin1990: Looks like most of the content in this recipe is duplicated with the rcu-service recipe. Perhaps we can explore and see if we can factor out those common content into a bb include file/class? This would helps in future maintenance and also reduce the risk of updating one recipe but forget to update the other.

Maybe we can structure our recipe like this?
recipe-core/rcu-service/rcu-service_1.0.bb
recipe-core/rcu-service/rcu-service-cpp-test-client_1.0.bb
recipe-core/rcu-service/rcu-service-common.inc

HatsyRei: That's a really good idea. I'll experiment around with creating our own inc file and see if things work as expected.

karhin1990: Thanks! If it takes you more than 3 hrs then don't do it. We can create an improvement task to track it and do it when we have time.

karhin1990 (approved):
Nice work!!

PR #9, opened 2021-03-05
Add custom base device tree, configure vsftpd

HatsyRei:


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-2.0+ OR X11
+/*
+ * Copyright 2019 Toradex
```
karhin1990: Need to use NI

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,259 @@
+// SPDX-License-Identifier: GPL-2.0+ OR X11
+/*
+ * Copyright 2019 Toradex
```
karhin1990: Ditto

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-2.0+ OR X11
```
karhin1990: Not sure what this license mean, but I'm fine if this is what we should use

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright 2019 Toradex
+ */
+
+/dts-v1/;
+
+#include "imx8qxp-apalis.dtsi"
+#include "imx8qxp-apalis-smartracks.dtsi"
+
+/ {
+	model = "Toradex Apalis iMX8QXP/DX on SmartRacks Custom Board";
```
karhin1990: Nit: Remove 'DX' since it's not the variant we are using. i.e. "Toradex Apalis iMX8QXP on SmartRacks Custom Board"

recipes-daemons/vsftpd/vsftpd_%.bbappend
```diff
@@ -1,5 +1,14 @@
...
+
+    install -d ${D}/${localstatedir}/lib/ftp
+    chown -R nobody:nogroup ${D}/${localstatedir}/lib/ftp
+    chmod -R 775 ${D}/${localstatedir}/lib/ftp
+
+    install -d ${D}/${localstatedir}/lib/ftp/upload
+    chown -R ftp:ftp ${D}/${localstatedir}/lib/ftp/upload
+    chmod -R 777 ${D}/${localstatedir}/lib/ftp/upload
+
     install -m 644 ${WORKDIR}/vsftpd.conf ${D}${sysconfdir}
```
karhin1990, line 16: ${D}${sysconfdir}: are we missing / in between ${D} and ${sysconfdir} ?

recipes-daemons/vsftpd/files/vsftpd.conf
```diff
@@ -1,141 +1,23 @@
...
+# http://vsftpd.beasts.org/vsftpd_conf.html
 
-# run standalone
-listen=YES
-
-# Allow anonymous FTP? (Beware - allowed by default if you comment this out).
-anonymous_enable=NO
-#
-# Uncomment this to allow local users to log in.
 local_enable=YES
```
karhin1990, line 21: For each setting in this file, can you add a comment on "why" we use that specific value? E.g. why we want to set local_enable to yes, why we should set local_umask=022

It helps to keep track the history so that in future we can trace back why we set these value at the first place.

recipes-daemons/vsftpd/files/vsftpd.conf
```diff
@@ -1,141 +1,23 @@
-# Example config file /etc/vsftpd.conf
```
karhin1990, line 1: For each setting that has been commented out, I would recommend to keep it instead of remove it.  This helps us to understand what is the possible setting available and why we decided to comment it out.

karhin1990 (commented):
Looks like we are still using the previously created overlay. What is the reason to still use overlay even we have created a smartracks specific device tree?

recipes-daemons/vsftpd/vsftpd_%.bbappend
HatsyRei: I checked the variable sysconfdir in [bitbake.conf](https://github.com/openembedded/openembedded-core/blob/dunfell/meta/conf/bitbake.conf) and found out that the variables are always prefixed with a slash ( / ). This means that either ${D}${sysconfdir} or ${D}/${sysconfdir} works since consecutive slashes in the path equate to one slash ([ref](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271)). I should probably standardize and remove the slash between D and localstatedir.

HatsyRei:
> Looks like we are still using the previously created overlay. What is the reason to still use overlay even we have created a smartracks specific device tree?

Currently, changes to the device tree are still done using overlays. We plan to add the base device tree first, and then migrate the changes made using the overlay (e.g. pinmux, disabling/enabling buses) over into the base device tree. The overlays will then be solely used for customizing platform devices.

recipes-daemons/vsftpd/vsftpd_%.bbappend
karhin1990: Good to know! I'm going to resolve this conversation. Thanks!

karhin1990 (commented):
Overall looks good, will wait for the udev change before Approving this PR

recipes-core/udev/udev-extraconf_%.bbappend
```diff
@@ -0,0 +1,13 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
+
+SRC_URI_append = " file://vsftpd-sd.rules \
+                   file://vsftpd-sd.sh \
+"
+
+do_install_append() {
+    install -d ${D}${sysconfdir}/udev/rules.d
+    install -m 0644 ${WORKDIR}/vsftpd-sd.rules    ${D}${sysconfdir}/udev/rules.d/vsftpd-sd.rules
```
karhin1990: nit: Additional space before ${D}

recipes-core/udev/files/vsftpd-sd.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+# Called from udev
+#
+# Starts/stops vsftpd.service on SD card hotplug
+
+
+if [ "$ACTION" = "add" ]; then
+	# Start vsftpd.service
+	mount /dev/mmcblk1p1 /var/lib/ftp/upload
```
karhin1990, line 10: Is `mount` return any status code? If yes then would be good to check it and skip the subsequent line (systemctl start vsftpd) if error happen. If error happen, also good to log it.

recipes-core/udev/files/vsftpd-sd.sh
```diff
@@ -0,0 +1,20 @@
...
+	systemctl start vsftpd
```
karhin1990: Will be good to log the status return from this line, same goes to "systemctl stop vsftpd" in line 15

recipes-daemons/vsftpd/vsftpd_%.bbappend
```diff
@@ -1,5 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
+SYSTEMD_AUTO_ENABLE_${PN} = "disable"
```
karhin1990, line 4: Can you add a comment saying the service starting/stopping is handled by udev rule?

conf/machine/apalis-imx8x-smartracks.conf
```diff
@@ -68,6 +68,6 @@ IMAGE_FSTYPES += "teziimg"
 UBOOT_BINARY_TEZI_EMMC = "imx-boot"
 OFFSET_BOOTROM_PAYLOAD = "0"
 
-TEZI_EXTERNAL_KERNEL_DEVICETREE_BOOT = "smartracks_overlay.dtbo"
```
karhin1990, line 71: Are we going to remove/clean up smartracks_overlay.dts, template_overlay.dts and device-tree-overlays_git.bb?

karhin1990 (commented):
Looks really good! Only few minor comments then we are good to go!

conf/machine/apalis-imx8x-smartracks.conf
HatsyRei: We can remove them if we do not foresee using them in near future. I can document those overlays stuff somewhere so that we can refer to it if we need to use overlays again sometime.

PR #10, opened 2021-03-16
Add rcu-service-python-test-client.bb file

tanchunhau:


recipes-core/rcu-service/rcu-service-python-test-client.bb
```diff
@@ -0,0 +1,20 @@
+SUMMARY = "RCU Service Python Test Client"
+DESCRIPTION = "Recipe to install RCU Service Python Test Client into RCU image"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
```
HatsyRei: Do we need to specify SRC_URI and SRCREV here? Just for sake of completeness. I understand the test client code will only be in rcu-service once the latest PR goes through.

recipes-core/rcu-service/rcu-service-python-test-client.bb
```diff
@@ -0,0 +1,20 @@
+SUMMARY = "RCU Service Python Test Client"
```
karhin1990: Need to specify revision in the file name: rcu-service-python-test-client_1.0.bb

recipes-core/rcu-service/rcu-service-python-test-client.bb
```diff
@@ -0,0 +1,20 @@
...
+DESCRIPTION = "Recipe to install RCU Service Python Test Client into RCU image"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+DEPENDS += " python3-grpcio-tools-native"
+inherit python3native
+
+do_compile()
+{
+	python3 -m grpc_tools.protoc -I${S} --python_out=${S}/tests --grpc_python_out=${S}/tests ${S}/rcu-service.proto
```
karhin1990: Since there is also cpp source in the ${S}/tests folder, it will be good to have a dedicated folder specifically for python test client. I.e. ${S}/tests/python

recipes-core/rcu-service/rcu-service-python-test-client.bb
tanchunhau: will add, I forget to add this, thanks for capture this

karhin1990: I think we need to define SRC_URI and SRCREV in a single file, then have all the recipes within recipes-core/rcu-service to include it, so that all of them are actually referring to the same source version. What do you think?

recipes-core/rcu-service/rcu-service-python-test-client.bb
HatsyRei: All paths referring to python script for RCU Service test client should now point to ${S}/tests/python. Will update the RCU Service PR so that the recipe can find the python scripts where it expects to.

karhin1990 (approved):
Nice!

PR #11, opened 2021-03-23
Update device tree and udev rules

HatsyRei:
DT: Add support for 2nd ethernet port
udev: Modify and include Toradex udev rules for custom apalis-smartracks machine


recipes-core/udev/files/apalis-imx8x-smartracks/99-toradex.rules
```diff
@@ -0,0 +1,18 @@
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="5a800000.i2c", SYMLINK+="I2C0"
```
karhin1990: Since we have add in our own implementation, let's call it 99-ni.rules~ 
You can put a comment on top of the file to say the content of this file is mainly copied from 99-toradex.rules

recipes-core/udev/files/apalis-imx8x-smartracks/99-toradex.rules
```diff
@@ -0,0 +1,18 @@
...
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="5a820000.i2c", SYMLINK+="I2C4"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="58226000.i2c", SYMLINK+="I2C5"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="56226000.i2c", SYMLINK+="I2C6"
+ACTION=="add", ATTRS{iomem_base}=="0x5A070010", SYMLINK+="UART1"
+ACTION=="add", ATTRS{iomem_base}=="0x5A060010", SYMLINK+="UART2"
+ACTION=="add", ATTRS{iomem_base}=="0x5A080010", SYMLINK+="UART3"
+ACTION=="add", ATTRS{iomem_base}=="0x5A090010", SYMLINK+="UART4"
+ACTION=="add", KERNELS=="watchdog", SYMLINK+="apalis-watchdog"
+ACTION=="add", SUBSYSTEM=="spidev", KERNELS=="5a000000.spi", SYMLINK+="apalis-spi1-cs%n"
+ACTION=="add", SUBSYSTEM=="spidev", KERNELS=="5a020000.spi", SYMLINK+="apalis-spi2-cs%n"
```
karhin1990: Since we don't support SPI, you can either remove these 2 lines or comment it out

recipes-core/udev/files/apalis-imx8x-smartracks/toradex-adc.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
```
karhin1990: Similarly, call it ni-adc.sh, and put a comment saying this file content is copied from toradex-adc.sh

recipes-core/udev/udev-toradex-rules.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
```
karhin1990: Question:
- how does toradex recipe selectively install udev rules based on the board model? I have no clue how it do it after read through its recipe (https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/udev-toradex-rules.bb?h=dunfell-5.x.y)
- .bbapend vs .bb: Looks like we creating our own udev rules, instead of 'appending' to toradex rules right?

HatsyRei: Yeah, honestly speaking SRC_URI automatically selecting machine-specific files is news to me too. I'm not sure exactly how it works, but I managed to find this:
https://stackoverflow.com/a/49281083
https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#var-SRC_URI_OVERRIDES_PACKAGE_ARCH

Since we are still using the udev recipe in meta-toradex-bsp-common to do_install those rules, appending seems to be the way to go. All we need to do is make sure udev-toradex-rules can "see" our machine-specific udev rules.

Due to how this works, renaming these rules with custom names will require our bbappend file to specify extra SRC_URI content and do_install steps to install them into the system, since OE considers them different files. E.g. If we call it 99-ni.rules instead of 99-toradex.rules, we will have to specify 99-ni.rules in SRC_URI, then make a do_install_append to install it. Another way will be to just ditch appending to udev-toradex-rules.bb and create something like udev-ni-rules.bb where we install our own set of udev rules, but keep in mind placeholder files from udev-toradex-rules will still be installed.

karhin1990: Ah I see!! Thank you so much for explaining, now I get the entire picture! 

Since this is the case, I'm fine with either appending udev-toradex-rules (your current implementation) or create a new udev-ni-rules.bb. The latter seems cleaner and more independent (because we don't need to worry one day Toradex change its .rules filename) but I don't see any major issue with the former as well. You can make the call.  :)

recipes-core/udev/files/apalis-imx8x-smartracks/toradex-adc.sh
HatsyRei: Will be removing this file since we don't really have any use for creating adc symlinks.

recipes-core/udev/udev-toradex-rules.bbappend
HatsyRei: Created a new udev-ni-rules recipe to install 99-ni.rules. Removed adc rules since we have no use for those symlinks.

karhin1990 (approved):

PR #21, opened 2021-04-19
Updates to base image and RCU Service Python client recipe

HatsyRei:
Packages cifs-utils and keyutils required to mount \\argo server to access SmartRacks RCU package feed.

### DIFF
--- a/recipes-images/images/packagegroup-ni-cli.bb
+++ b/recipes-images/images/packagegroup-ni-cli.bb
@@ -23,6 +23,8 @@ RRECOMMENDS_packagegroup-base-ni-cli = "\
     coreutils \
     rcu-service \
     udev-ni-rules \
+    keyutils \
+    cifs-utils \
 "
 
 SUMMARY_packagegroup-devel-ni-cli = "Tools useful during SmartRacks RCU development"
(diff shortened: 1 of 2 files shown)


recipes-images/images/packagegroup-ni-cli.bb
```diff
@@ -23,6 +23,8 @@ RRECOMMENDS_packagegroup-base-ni-cli = "\
     coreutils \
     rcu-service \
     udev-ni-rules \
+    keyutils \
+    cifs-utils \
```
karhin1990, line 27: Can you share a little bit why we are adding these 2 packages? Perhaps add a comment on the PR description will be helpful as it can serve as future references as well. Thanks!

karhin1990 (approved):

recipes-images/images/packagegroup-ni-cli.bb
tanchunhau: this is for auto-mount (login credential) to NI-Network directory, so HW team can download/install package feed.
But I am not sure should it be installed inside image, maybe we can discuss offline later.

HatsyRei: Yeah, we actually discussed this back in March but I forgotten to add them to the image. It only struck me yesterday when I tried to run the setupfeed script and realized we don't have these packages to enable mounting the argo server on the RCU. I checked our Teams chat and I think we agreed to put them into the image:

> I think that's fine to preinstall, since they are the bootstrap package to enable use of the feed to install further packages.

karhin1990: Thanks for explanation! Resolving this issue

PR #24, opened 2021-04-26
DT: Add support for analog I2C mux on fanout board

HatsyRei:


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -46,6 +46,36 @@
...
+    
+    /* I2C Analog Mux attached to Apalis I2C3 */
+    i2cmux {
+        compatible = "i2c-mux-gpio";
+        #address-cells = <1>;
+        #size-cells = <0>;
+        mux-gpios = <&lsio_gpio3 21 0 &lsio_gpio3 22 0>;
+        i2c-parent = <&i2c3>;
+
+        i2c@0 {
```
tanchunhau, line 58: I have some questions, what will be the name we saw in `ls /dev `?
i2c@0 become ?
i2c@1 become ?
Is the new generated name is always fix, or it is random (after build new image) ?
For example,
/dev/i2c_mux_ch1 is always  = i2c3.i2c@0 ?

HatsyRei: Same case for muxed i2c bus as is for other physical i2c buses in the DT: The enumerated bus number in /dev may change. We will have to create symlinks in udev.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -46,6 +46,36 @@
         /* Apalis USBH_EN */
         gpio = <&lsio_gpio4 4 GPIO_ACTIVE_HIGH>;
     };
+    
+    /* I2C Analog Mux attached to Apalis I2C3 */
+    i2cmux {
+        compatible = "i2c-mux-gpio";
+        #address-cells = <1>;
+        #size-cells = <0>;
+        mux-gpios = <&lsio_gpio0 24 0 &lsio_gpio3 15 0>;
```
karhin1990: Just to double check: It seems like we need to list the GPIO that control the least significant bit of the mux first. 

In that sense, means `&lsio_gpio0 24 0` is connected to A0 of the mux; while `&lsio_gpio3 15 0` is connected to A1 of the mux, right?

karhin1990 (commented):
Solution looks good to me, will be approving it once the udev rule change is uploaded.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
tanchunhau: okay.
maybe I learn from the udev PR later. thank you.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Ah... you are right. I don't think the carrier pinout gave sufficient information for us to determine which one should be specified first/last for now. These two pins were specified to be used for "temp sensor selection", but makes no mention on which should go to A0 and which to A1. On the fanout board schematics, A0 and A1 were left unconnected to anything. In the fanout board user manual, the instructions were to connect A0 and A1 to GPIO, but makes no mention of which specific pin to connect to. I have yet to check if the Rev A schematics to see if anything is mentioned for these pins.

karhin1990: Alright. Just want to make sure we assign it correctly so that we can set the channel name properly (the `reg` field) which aligns to the channel number assigned in the datasheet.

Since now we are using the fanout board, we just need to make sure `&lsio_gpio0 24 0` is connected to A0 of the mux; while `&lsio_gpio3 15 0` is connected to A1 of the mux, then that's good enough.

recipes-core/udev/files/apalis-imx8x-smartracks/99-ni.rules
```diff
@@ -6,6 +6,10 @@ ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="5a830000.i2c", SYMLINK+="I2C3
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="5a820000.i2c", SYMLINK+="I2C4"
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="58226000.i2c", SYMLINK+="I2C5"
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="56226000.i2c", SYMLINK+="I2C6"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="i2c-[0-9]*-mux (chan_id 0)", SYMLINK+="I2C3_0"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="i2c-[0-9]*-mux (chan_id 1)", SYMLINK+="I2C3_1"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="i2c-[0-9]*-mux (chan_id 2)", SYMLINK+="I2C3_2"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTRS{name}=="i2c-[0-9]*-mux (chan_id 3)", SYMLINK+="I2C3_3"
```
karhin1990: Sorry for nit-picking: I was thinking would it be clearer if we call it "I2C3_MUX1_CH0"? That's more expandable in case in future there are more than 1 mux sitting on the same parent bus..

HatsyRei: Updated symlink name accordingly.

karhin1990 (approved):
Nice!

PR #27, opened 2021-05-12
add fan controller (MAX31785) into smartracks.cfg

tanchunhau:
Add FanController chip (MAX31785)

For device tree, I follow this : (no longer valid after offline discussion)
https://www.kernel.org/doc/Documentation/devicetree/bindings/hwmon/max31785.txt

For smartracks.cfg:
I use linux-toradex menuconfig to generate.


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -200,6 +200,11 @@
     pinctrl-0 = <&pinctrl_lpi2c4>;
     clock-frequency = <100000>;
     status = "okay";
+
+    fans@52 {
```
tanchunhau: In actual connection, there is maximum 3 fan ctrl chip (different chip address). Please refer to image below.

Hi @HatsyRei ,
What is your decision for detect device ?
1.) Create a supper-set in device tree (then I need to add another two in this file)
2.) Dont add in device tree, and let c++ code load it during runtime.
3.) Any others ?

Please share your opinion, thank you.

image

HatsyRei: @karhin1990 Since we plan on going with the 2) option for thermal sensors, I imagine we will do the same for fan chip in this case?

tanchunhau: discussed offline, go for method 2.)

Dont add in device tree, and let c++ code load it during runtime.

karhin1990 (approved):

HatsyRei (approved):
Thanks for fixing the stray whitespaces :D

PR #28, opened 2021-05-19
Mux CAN bus pins to GPIO in Device Tree (Pin 12,14,16,18)

tanchunhau:
I am learning how to add gpio pin,
below are the details:
 
MXM3_18 is for status LED (**Pin Green color LED**)

It is from  (already disabled)
```
/* Apalis CAN2 */
&flexcan3 {
    /* define the following property to disable CAN-FD mode */
    /* disable-fd-mode; */
    status = "disabled";
};
```
image

and from the upper layer file :

image

### DIFF
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -520,12 +520,6 @@
                 IMX8QXP_UART0_RX_ADMA_UART0_RX 0x06000020 /* MXM3 132 */
             >;
         };
-
-        pinctrl_flexcan3: flexcan3grp {
-            fsl,pins = <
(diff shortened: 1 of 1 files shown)


karhin1990 (commented):
Since we are not using CAN anymore, can you also help to release the other GPIO pins that was mux-ed to CAN bus previously? I believe they are pin 12, 14 and 16 (based on pinout designer)

karhin1990:
Also, I would make the commit message more descriptive (it will be helpful for us to find back a specific commit from the history log in future), such as: "Configure pin mxm3_18 as GPIO in Device Tree", or "Mux CAN bus pins to GPIO in Device Tree"

HatsyRei (approved):
Looks good to me!

karhin1990 (approved):

PR #29, opened 2021-06-14
Fix 48-bit IO Expander driver patch

HatsyRei:
Fix bug in pcf857x driver patch involving 48-bit IO expanders, i.e. pi4ioe5v96248.
Bug caused by use of unsigned int (32-bit) to read default pin state and apply to 48-bit IO expander.
This results in state 0 being written to the last two banks (16 IOs) regardless of applied default line values.

The updated driver containing these fixes can be found here:
https://github.com/HatsyRei/diodes/blob/master/gpio-diodes.c

### DIFF
--- a/recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
+++ b/recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
@@ -1,5 +1,5 @@
---- a/drivers/gpio/gpio-pcf857x.c	2021-06-10 04:00:37.797098202 -0700
-+++ b/drivers/gpio/gpio-pcf857x.c	2021-06-14 05:50:13.961515083 -0700
+--- a.gpio-pcf857x.c	2021-06-10 04:00:37.797098202 -0700
++++ b.gpio-pcf857x.c	2021-06-16 01:06:23.203382752 -0700
 @@ -17,6 +17,7 @@
  #include <linux/of_device.h>
  #include <linux/slab.h>
@@ -155,7 +155,7 @@
  	struct pcf857x  *gpio = data;
 -	unsigned long change, i, status;
 +	u64 value;
-+	unsigned long change;
++	unsigned long change; // unsigned long is 8 bytes wide on ARM64
 +	int status, i;
  
 -	status = gpio->read(gpio->client);
--- a/recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
+++ b/recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
@@ -1,5 +1,5 @@
 --- a/drivers/gpio/gpio-pcf857x.c	2021-06-10 04:00:37.797098202 -0700
-+++ b/drivers/gpio/gpio-pcf857x.c	2021-06-16 01:06:23.203382752 -0700
++++ b/drivers/gpio/gpio-pcf857x.c	2021-06-17 07:51:18.812157083 -0700
 @@ -17,6 +17,7 @@
  #include <linux/of_device.h>
  #include <linux/slab.h>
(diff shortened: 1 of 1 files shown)


HatsyRei:
@karhin1990 I tried changing the unsigned long variable we found in pcf857x_irq to u64, but seems like 'find_first_bit' requires its arg to be unsigned long pointer and not u64 pointer.

https://elixir.bootlin.com/linux/latest/source/include/asm-generic/bitops/find.h#L59

I haven't noticed any buggy behavior caused by this, could be due to how we don't use interrupts on the IO expander yet. Perhaps we can note down this concern for now.

karhin1990:
> @karhin1990 I tried changing the unsigned long variable we found in pcf857x_irq to u64, but seems like 'find_first_bit' requires its arg to be unsigned long pointer and not u64 pointer.
> 
> https://elixir.bootlin.com/linux/latest/source/include/asm-generic/bitops/find.h#L59
> 
> I haven't noticed any buggy behavior caused by this, could be due to how we don't use interrupts on the IO expander yet. Perhaps we can note down this concern for now.

Yea.. I think we should be safe since we are not using interrupt. Perhaps it's worth to add a comment there so that in future if we use interrupt, we know that is the place we need to fix. 

Anyways, if it's not too much trouble, can you try runs an app in the SOM to return sizeof(unsigned long)? I'm pretty curious how long this type actually is..

tanchunhau (approved):
sui !

recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
```diff
@@ -1,5 +1,5 @@
---- a/drivers/gpio/gpio-pcf857x.c	2021-05-02 05:15:14.347074972 -0700
-+++ b/drivers/gpio/gpio-pcf857x.c	2021-05-02 05:14:56.523219426 -0700
+--- a/drivers/gpio/gpio-pcf857x.c	2021-06-10 04:00:37.797098202 -0700
```
tanchunhau, line 1: I try to understand:
I saw from your link : https://github.com/HatsyRei/diodes/blob/master/gpio-diodes.c

In line 383: (new)
gpio->status = gpio->read(gpio->client);

Why cannot use back the original (use latched value) after change the "latch" to u64 ?
gpio->status = gpio->out; (original from maquefel)

Regarding "long" type, I think don't change it to u64 is better if it is be used for address.
int find_first_bit (const unsigned long * addr, unsigned size);
PS : for our case, 'long' is 8 Byte (64 bit) because we are 64bit Arm system.

But I dont see the related line ? it is because you remove the irq related already ?

tanchunhau (approved):

recipes-kernel/linux/files/add-support-for-diodes-pi4ioe5v96248.patch
HatsyRei: gpio->status = gpio->out;

is outdated driver code I think. Can be found in kernel versions around 4.14 to 4.19
https://git.toradex.com/cgit/linux-toradex.git/tree/drivers/gpio/gpio-pcf857x.c?h=toradex_4.14-2.3.x-imx#n360

It was changed to gpio->status = gpio->read(gpio->client); in later versions of pcf857x:
https://git.toradex.com/cgit/linux-toradex.git/tree/drivers/gpio/gpio-pcf857x.c?h=toradex_5.4-2.3.x-imx#n335

As for find_first_bit, it is called by for_each_set_bit in the driver code, which is a macro defined here: 
https://elixir.bootlin.com/linux/latest/source/include/linux/bitops.h#L34

You can find this line in the modified driver:
for_each_set_bit(i, &change, gpio->chip.ngpio)

https://github.com/HatsyRei/diodes/blob/master/gpio-diodes.c#L226

tanchunhau: i see, thanks you.

HatsyRei: @karhin1990 sizeof(unsigned long) on the SOM gave me 8 (bytes) btw, so 64 bit. :D

karhin1990: Put a comment there, then case closed!  ;)

HatsyRei: Added a comment in the gpio-diodes.c file. Should I also regen the patch so that the comment is captured in our patch file?

karhin1990: yaya, that will be helpful..

HatsyRei: Done.

karhin1990 (commented):
Current changes looks good to me. I will approve it after you've added "lines-initial-states" in the DT and tested it can be consumed correctly by `of_property_read_u64`

HatsyRei:
@karhin1990 Aight. I've dev tested "lines-initial-states" earlier with alternating bitmasks (101010101....) and it appears to work fine. I'm introducing the "lines-initial-states" property into the DT in this PR:

https://github.com/ni/meta-smartracks/pull/32

karhin1990:
Great. Last thing before approving this PR would be the `of_read_property_32` and `of_read_property_64` selection based on `gpio->chip.ngpio`

karhin1990 (approved):
Perfecto!

PR #30, opened 2021-06-15
DT: Adjustments to suit carrier board schematics

HatsyRei:
- Disabled SPI1 to free up MXM3 221, 223, 225, 227 for GPIO
- Move pi4ioe5v IO expander from bus I2C1 to I2C4
- Move analog I2C mux parent from bus I2C3 to I2C5
- Add PCA9674 8-bit IO expander on PEP to I2C3


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -47,14 +47,14 @@
         gpio = <&lsio_gpio4 4 GPIO_ACTIVE_HIGH>;
     };
 
-    /* I2C Analog Mux attached to Apalis I2C3 */
+    /* I2C Analog Mux attached to Apalis I2C5 */
```
karhin1990, line 50: Need to update our udev rule, tI2CManager (cpp and .h) and rcu-service.proto as well to update the bus naming (e.g. `I2C3_MUX1_CH0`)

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -200,6 +184,22 @@
     pinctrl-0 = <&pinctrl_lpi2c4>;
     clock-frequency = <100000>;
     status = "okay";
+    
+    /* 48-bit IO expander */
```
karhin1990, line 198: There is a piece of code in tGPIOManager to open Chip 10 for the IO Expander. Not sure will the chip number changed after we relocated the IO Expander. Might want to check it out.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -308,8 +308,9 @@
 };
 
 /* Apalis SPI1 */
+/* Disabled to free up MXM3 221, 223, 225, 227 for GPIO */
 &lpspi0 {
-    status = "okay";
+    status = "disabled";
```
karhin1990, line 324: Regarding SPI 2, after we disabled it looks like we can also free up pin 204 for GPIO right? I was checking it with pinout designer

karhin1990 (commented):
Nice work!
(FYI: You can find Rev A carrier board pin mapping [here](https://nio365.sharepoint.com/:u:/r/sites/SmartRack/Shared%20Documents/4_Software/Pin%20Mapping/Rev%20A%20Carrier%20Board/Apalis_iMX8QXP_ECC_RevACarrierBoard.xml?csf=1&web=1&e=kE68JN))

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: tAteCoreCntApiServiceImpl.cpp as well

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: MXM3 204 seems to be GPIO by default:
https://git.toradex.com/cgit/linux-toradex.git/tree/arch/arm64/boot/dts/freescale/imx8qxp-apalis.dtsi?h=toradex_5.4-2.3.x-imx#n604

though seems like we do not have this pin programmed into rcu-service yet.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Tested by hardware engineer previously, we saw no change in the gpiochip number. Not too sure what happens if we add a second IO expander though. My understanding is that the built-in GPIO controllers get initialized first and then IO expanders. For IO expanders, the gpiochip number depends on the order of initialization, which follows the I2C bus init order. 

E.g. We would expect the chip number to become 11 if we add another IO expander on bus 3 (before bus 4).

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: Yea.. I guess we were reserving it as the Chip Select pin for spi2.. Can you help to add it into rcu-service as well (as part of https://github.com/ni/rcu-service/pull/38)?

HatsyRei: Yep, will do. I'm looking at the Rev A pinout .xml you linked. Will make sure that we have all the GPIOs listed in there. 204 was part of it.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: Hmm.. That's bad.. perhaps we need to look for a way to get the chip number dynamically rather than always hardcode it to 10 in tGPIOManager. 

Let's add a Todo to tGPIOManager (where it hard code the number 10) as well as create an Azdo work item to track it, we might want to revisit it and tackle it later (unless you got an idea in mind that can fix it within 30 minutes =P)

HatsyRei: Seems like the function we are using rn is gpiod_chip_open_by_number
https://libgpiod-dlang.dpldocs.info/gpiod.gpiod_chip_open_by_number.html

We can instead use gpiod_chip_open_by_label, which, looking at its implementation in gpiodetect.c, appears to be exactly what we are looking for. I see stuff like "5d080000.gpio", "5d090000.gpio", etc being printed as labels when running gpiodetect on the SOM, which should be more fixed than the enumerated chip number. 
https://libgpiod-dlang.dpldocs.info/gpiod.gpiod_chip_open_by_label.html

Changing our existing implementation can take some time though... We should probably create a work item like you said. There's still a problem with these labels, seeing that the onboard IO expander was labelled "16-0020" (16 being the enumerated i2c bus number) and our PI4IOE5V labelled as "pi4ioe5v96248". We will have to figure out whether we can make them more specific and independent.

karhin1990: Good findings! Yea I agree that let's create a Todo to capture it~

HatsyRei: Added this as a technical debt here:
https://dev.azure.com/ni/DevCentral/_backlogs/backlog/Smart%20Racks%20Software/Features/?workitem=1500033

karhin1990: I converted it to a Task item and added it to this user story (https://dev.azure.com/ni/DevCentral/_workitems/edit/1407841) since we used to consolidate Todos there.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Seems like 204 is the only one we missed out. Added to our DT and also updated in corresponding rcu-service PR.

HatsyRei:
DT: Added PCA9674 8-bit IO expander on PEP to I2C3.

karhin1990 (approved):
Good job!

PR #32, opened 2021-06-17
Add labels and lines-initial-states to IO Expander nodes

HatsyRei:
Add labels for IO Expanders. Labels will be used to uniquely identify IO expander chips in rcu-service.
Add lines-initial-states for pi4ioe5v96248 IO expander, initializes IO1_4 and IO1_5 to output 0.

### DIFF
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -181,6 +181,7 @@
         reg = <0x20>;
         gpio-controller;
         #gpio-cells = <2>;
+        label = "pca9674-5a830000.gpio";
         gpio-line-names =
             "IOXPD2_IO0_0", "IOXPD2_IO0_1", "IOXPD2_IO0_2", "IOXPD2_IO0_3", "IOXPD2_IO0_4", "IOXPD2_IO0_5", "IOXPD2_IO0_6", "IOXPD2_IO0_7";
     };
@@ -202,6 +203,7 @@
         gpio-controller;
         #gpio-cells = <2>;
         npgio=<48>;
+        label = "pi4ioe5v96248-5a820000.gpio";
         gpio-line-names =
             "IOXPD1_IO0_0", "IOXPD1_IO0_1", "IOXPD1_IO0_2", "IOXPD1_IO0_3", "IOXPD1_IO0_4", "IOXPD1_IO0_5", "IOXPD1_IO0_6", "IOXPD1_IO0_7",
             "IOXPD1_IO1_0", "IOXPD1_IO1_1", "IOXPD1_IO1_2", "IOXPD1_IO1_3", "IOXPD1_IO1_4", "IOXPD1_IO1_5", "IOXPD1_IO1_6", "IOXPD1_IO1_7",
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -204,6 +204,7 @@
         #gpio-cells = <2>;
         npgio=<48>;
         label = "pi4ioe5v96248-5a820000.gpio";
+        lines-initial-states = <0xC0000 0x0>;
         gpio-line-names =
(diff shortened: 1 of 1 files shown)


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -202,6 +203,8 @@
         gpio-controller;
         #gpio-cells = <2>;
         npgio=<48>;
+        label = "pi4ioe5v96248-5a820000.gpio";
+        lines-initial-states = <0xC0000 0x0>;
```
karhin1990: Isn't it we want all bits to be zero, except IO1_5 and IO1_6? IO1_5 and IO1_6 seems to be bit 45 and bit 46 of the 64bit data

Also, it's good to put a comment that says "IO1_5" and "IO1_6" has to be in LOW state by default. Hence they need to set to "0" in lines-initial-states

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -181,6 +181,7 @@
         reg = <0x20>;
         gpio-controller;
         #gpio-cells = <2>;
+        label = "pca9674-5a830000.gpio";
```
karhin1990, line 184: can we add the lines-initial-states for this one as well?

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Yep. @karhin1990 correct me if I am wrong. Just to keep a record of our discussion here:

1) A 64-bit lines-initial-states property store two 32-bit values consecutively, i.e. For a 64-bit value 0xFFFFFFFFAAAAAAAA
lines-initial-states = <0xFFFFFFFF 0xAAAAAAAA>

2) As for the what each bit in the bitmask represents, IO0_0 is LSB (bit at rightmost corner), and as we move left from LSB, we approach IO5_7 at the 48th bit. Everything between the 48th bit to MSB (bit at leftmost corner) does not represent anything (since IO expander is 48 bit only).

3) By having lines-initial-states of <0x00000000 0x00006000>, we have bitmask of:

00000000 00000000 00000000 00000000 00000000 00000000 01100000 00000000

Where bit 1s target IO1_5 and IO1_6.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Aight. I'll add them once I get a confirmation of the correct initial states.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: That's really good information, Yan Huan! Can I get your help to document it in the dtsi as a comment? It's a nightmare to reverse engineer the code and hope we won't fall into the same problem again in the future :)
A lengthy comment sometimes can saves us time. 

It's also good to have a brief comment to mention `lines-initial-states` is "active low".

HatsyRei: Added comments prior to lines-initial-states property in 48-bit IO expander node. 
Added lines-initial-states property to PEP CCA 8-bit IO expander node, according to RCU pinout document.

From the document, PEP CCA 8-bit IO expander IO0 and IO1 needs to be output LOW, so the bitmask should be 00000011, i.e. 0x03. @karhin1990 amirite? Just asking in case I messed up again :D

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Added.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: Looks like we are good~

karhin1990 (approved):
Here we go~

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
HatsyRei: Will dev test this before merging (at least the open by label part). Traumatized d :D

HatsyRei: Weee... It is working. Merging now.

PR #35, opened 2021-06-25
change initial state of I2C3

NIwkoe:
Hardware engineers reworked PEP FAN and PEP AUX to be active low so they have to be default high. Hence, setting I2C3 default to "output high" by setting bit 0 and bit 1 of the bit mask to 0.


HatsyRei (approved):

karhin1990:
The PR title is incorrect, you are changing the initial state of pins on IO Expander, not the I2C3 bus/port. Since you have merged, nothing much we can do. Just be aware on future PR.

NIwkoe:
> The PR title is incorrect, you are changing the initial state of pins on IO Expander, not the I2C3 bus/port. Since you have merged, nothing much we can do. Just be aware on future PR.

Ok, my bad for bad title. Will be cautious next time.

PR #36, opened 2021-06-28
Add rcu-version template files

HatsyRei:
PR adds an empty rcu-version file which is appended to /etc directory. This file should be modified by pipeline, which inserts the build number, allowing users to cat /etc/rcu-version to know the version of which rcu-image they are running.

### DIFF
--- a/recipes-core/rcu-version/rcu-version.bb
+++ b/recipes-core/rcu-version/rcu-version.bb
@@ -1,0 +1,11 @@
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = "\
+    file://rcu-version \
+"
+
+do_install () {
+    install -d ${D}${sysconfdir}
+    install -m 0644 ${WORKDIR}/rcu-version ${D}${sysconfdir}
+}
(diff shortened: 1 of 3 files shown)


recipes-core/rcu-version/rcu-version.bb
```diff
@@ -0,0 +1,11 @@
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = "\
+    file://rcu-version \
```
karhin1990: Would be helpful to name the bb and the version file as "rcu-**image**-version", so that it's clear that it representing the version for RCU Image instead of rcu-service version or the RCU hardware version.

karhin1990 (approved):

recipes-core/rcu-version/rcu-version.bb
HatsyRei: Altered all instances to rcu-image-version. Also added rcu-image-version to our packagegroup so that it is built when building smartracks-minimal-image.

karhin1990 (approved):
Nice! Was wondering if we should put file extension for the rcu-image-version file, like rcu-image-version.cfg, so that it looks clearer that it's a file rather than an executable.

Anyways, this is not a big deal, I will let you to make the call :)

PR #40, opened 2021-08-06
Fix Apalis iMX8X SPI2 pins MXM3 number mismatch

HatsyRei:


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -548,6 +548,47 @@
     };
 };
 
+/* Fix Apalis iMX8X SPI2 pins MXM3 number mismatch */
+/*  - MXM3_233 corresponds to GPIO chip 1 line 0   */
+/*  - MXM3_231 corresponds to GPIO chip 1 line 1   */
+/*  - MXM3_229 corresponds to GPIO chip 1 line 2   */
+/*  - MXM3_235 corresponds to GPIO chip 1 line 3   */
+/*  Ref: AzDO 1565419                              */
+&lsio_gpio1 {
```
karhin1990, line 558: Is the purpose of introducing this structure here to overwrite the similar struct that defined on imx8qxp-apalis.dtsi?
If yes, probably adding a comment to describe this would be helpful.

karhin1990 (approved):

PR #42, opened 2021-09-09
Enabling kernel config for ADS7830, PCA9552, AD7291, HDC1080 and BMP280

karhin1990:
Enable kernel driver for following I2C devices to support hardware testing and also future feature implementation:
- ADS7830: 8-bit ADC
- PCA9552: 16-bit LED
- AD729: 12-bit ADC
- HDC1080: Humidity sensor
- BMP280: Pressure sensor

Device tree implementation will be in separated PR.


nwilsonni:
For my understanding - Does this pull the Linux drivers into the build/image?  Then these devices will need to be enabled/added in the device tree to fully use the driver to interact with the hw.

karhin1990:
> For my understanding - Does this pull the Linux drivers into the build/image? Then these devices will need to be enabled/added in the device tree to fully use the driver to interact with the hw.

@nwilsonni - Yes, this PR enable the kernel driver support into the image (i.e. we build them into the image), but it won't load (to be exact, instantiate the kernel device object) automatically until we "tell" the OS that the corresponding device is there. Currently we are using 2 different ways to "tell" the OS the device is there:
1. Via Device Tree
2. Via runtime instantiation

For those device that we know it ALWAYS present in the rack (i.e. not customizable by integrator), we will hardcode them in the device tree. During RCU boot up, OS will decide whether or not to load the driver by looking for the presence of device from Device Tree. [Method 1]

For those device that might not always present (e.g. temp sensor board - some time it's there, something it isn't, based on integrator decision), our solution is to have rcu-service to read the config file, and then explicit ask the kernel to load the driver during runtime (instead of during boot time like Method 1). [Method 2]

Note that method 2 only support I2C devices (SPI and other buses are not supported), and it also depends on the driver requirement. Some I2C devices are not supporting Method 2, such as IO Expander.

karhin1990:
I'm going to submit this PR since there's no open comment.

PR #43, opened 2021-09-15
Adding PCA9552 16-bit LED driver to RCU Device Tree

karhin1990:
Justification:
- To ease hardware V&V, so that hardware engineer can control the LEDs via driver instead of have to perform raw I2C transaction

Description:
- Kernel driver was enabled in https://github.com/ni/meta-smartracks/pull/42
- Location of this LED driver: I2C Bus: 0x04; I2C Addr: 0x62
- DT binding guide: https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/leds/leds-pca955x.txt

Testing:
- Have not tested it as I don't have a board with me. Will work with hardware engineer to test it once the export is ready

### DIFF
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -226,6 +226,79 @@
         compatible = "dallas,ds3232";
         reg = <0x68>;
     };
+    
+    /* 16-bit LED driver */
+    pca9552: pca9552@62 {
+        compatible = "nxp,pca9552";
+        #address-cells = <1>;
+        #size-cells = <0>;
+        reg = <0x62>;
+
+        led@0 {
+            reg = <0>;
+            type = <PCA955X_TYPE_LED>;
+        };
+        led@1 {
+            reg = <1>;
+            type = <PCA955X_TYPE_LED>;
+        };
+        led@2 {
+            reg = <2>;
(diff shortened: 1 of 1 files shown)


HatsyRei (approved):

tanchunhau (approved):

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -226,6 +226,79 @@
...
+    
+    /* 16-bit LED driver */
+    pca9552: pca9552@62 {
+        compatible = "nxp,pca9552";
+        #address-cells = <1>;
+        #size-cells = <0>;
+        reg = <0x62>;
+
+        led@0 {
+            reg = <0>;
```
tanchunhau, line 238: Do we need to specify "linux,default-trigger" for this case ?

image

tanchunhau (approved):

NI-cphor (approved):
I will approve first because I am still not familiar to the code.
I am learning in the same time.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: TMK it's not needed, because the chip is used to control 7-segments LED.

nwilsonni (approved):

PR #45, opened 2021-11-03
Fix Diodes 48-bit GPIO Expander patch

HatsyRei:
Patch fix for Diodes GPIO Expander. This patch will be applicable if we bump git hashes in smartracks-manifest, here:
https://github.com/ni/smartracks-manifest/pull/10

Latest modified driver can be found here:
https://github.com/HatsyRei/diodes/blob/master/gpio-diodes.c


karhin1990 (approved):
Great catch and nice fix!

PR #46, opened 2021-12-23
Add smsc emc2305 driver to support new fan panel

NIwkoe:
linux-toradex's base image does not have native driver that supports smart racks' new fan panel that uses emc2303 (emc2305). A third-party driver patch was found online and looks usable ([source](https://patchwork.kernel.org/project/linux-hwmon/patch/20200928104326.40386-1-biwen.li@oss.nxp.com/#23650673)). It supports all functions that are referred in rcu-service's fan panel code. I amended the patch slightly to fit our need. Patch compiled fine as there was no compilation error and the emc2305.o.cmd can be found in the build artifacts. This driver is not tested yet though, creating this PR to go through a preliminary code review and get some feedbacks.


HatsyRei (commented):
This looks OK to me. If the patch applies correctly without warnings, I think my only concern would be to test that it works with actual hardware (and that dmesg says everything loads fine). I feel like we can leave out the documentation files, but will leave it to you to decide whether we should keep those.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2305.patch
```diff
@@ -0,0 +1,800 @@
+From 35bf96bc360b287ef63de3d14702320c704b58e4 Mon Sep 17 00:00:00 2001
```
karhin1990: Since this driver is written for smart rack project use only (instead of a generic upstream driver), I would recommend to reduce the scope of this driver to support EMC2303 only. This will make the code shorter, hence easier to review, maintain and debug. 

As such, let's name the driver as emc2303 instead of emc2305

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2305.patch
```diff
@@ -0,0 +1,800 @@
...
+@@ -0,0 +1,30 @@
++Kernel driver emc2305
++=====================
++
++Supported chips:
++  * SMSC EMC2305, EMC2303, EMC2302, EMC2301
++    Adresses scanned: I2C 0x2c, 0x2d, 0x2e, 0x2f, 0x4c, 0x4d
++    Prefixes: 'emc2305', 'emc2303', 'emc2302', 'emc2301'
++    Datasheet: Publicly available at the MICROCHIP website :
++        http://ww1.microchip.com/downloads/en/DeviceDoc/2305.pdf
```
karhin1990: This link doesn't seems correct

karhin1990 (commented):
Half way through. Will continue review it after you clean up those unnecessary codes (for EMC2301/5)

karhin1990:
You might also want to cater the settings/configurations that ME interested into the driver. Ee Yeng sent us a list of registers and its preferred value, you can refer to that.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++	.detect		= emc2303_detect,
++	.address_list	= i2c_addresses,
++};
++
++module_i2c_driver(emc2303_driver);
++
++MODULE_DESCRIPTION("SMSC EMC2303 hwmon driver");
++MODULE_LICENSE("GPL");
+-- 
+2.7.4
```
karhin1990, line 634: Just curious, what is the meaning of this?

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++	REG_TACH_READ_LOW = 0x3f,
++};
++
++#define SEL_FAN(fan, reg) (reg + fan * 0x10)
++
++/*
++ * Factor by equations [2] and [3] from data sheet; valid for fans where the
++ * number of edges equals (poles * 2 + 1).
++ */
++#define FAN_RPM_FACTOR 3932160
```
karhin1990: The equation in the datasheet you linked is 4-2 and 4-3. [2] and [3] are the one in old (SMSC) datasheet.

I don't understand what this statement means: "valid for fans where the number of edges equals (poles * 2 + 1)", maybe you can say "The RPM factor is only valid for fTACH = 32.765kHz, number of pole = 2 and number of edges measured = 5". Also, you need to check with ME to ensure the fTACH, number of pole and number of edges that we assume are correct, else the RPM factor will be invalid.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++
++struct emc2303_fan_data {
++	bool		enabled;
++	bool		rpm_control;
++	bool		valid;		/* registers are valid */
++	u8		poles;
++	u8		multiplier;
++	u16		tach;
++	u16		target;
++	unsigned long		last_updated;	/* in jiffies */
```
karhin1990: nit: spacing between `unsigned long` and `last_updated` is off

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++};
++
++struct emc2303_data {
++	struct i2c_client	*client;
++	const struct		attribute_group *groups[6];
```
karhin1990: The array size should be 4 (3 fan group + 1 NULL terminator), I believe

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
+From: wkoe <wilson.koe@ni.com>
+Date: Wed, 22 Dec 2021 23:17:08 -0800
+Subject: [PATCH] hwmon: add support for SMSC EMC2303
+
+Signed-off-by: wkoe <wilson.koe@ni.com>
+---
+ drivers/hwmon/Kconfig   |  10 +
+ drivers/hwmon/Makefile  |   1 +
+ drivers/hwmon/emc2303.c | 592 ++++++++++++++++++++++++++++++++++++++++++++++++
+ 3 files changed, 603 insertions(+)
```
karhin1990: Thinking further, I think it is make sense to also include the device tree binding documentation into this patch as well, so that in future people know how to define the DT binding for this driver.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++		data->groups[idx++] = &emc2303_fan3_group;
++	}
++	else if (status < 0)
++	{
++		dev_err(&client->dev, "Failed to read reg REG_PRODUCT_ID, status = %d\n", status);
++		return status;
++	}
++	else
++	{
++		dev_err(&client->dev, "Unknown device, status = %d\n", status);
```
karhin1990: if `status` >= 0, meaning `i2c_smbus_read_byte_data` actually successfully read the register and it fill `status` with the register content. Hence, the correct debug message should be: `dev_err(&client->dev, "Unknown device, REG_PRODUCT_ID register content = 0x%x\n", status);`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,647 @@
...
++		return status;
```
karhin1990: Should return `-ENODEV`

karhin1990 (commented):
Half way through, to be continued...

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2305.patch
NIwkoe: I've slimmed down the driver to only support emc2303 by removing codes related to configuring for 5 fans. I've also renamed all occurrence of emc2305 name to emc2303.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2305.patch
NIwkoe: Changed to a valid link.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Found it on google:
https://superuser.com/questions/1462869/git-format-patch-number-at-end-of-patch

Apparently it is a neat feature to tell what git version is used when running

`git format-patch`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Thanks for spotting this, the comment was outdated. I checked with ME's register map, they use
fTACH = 32kHz
no. of pole = 2
edges = 5

so applying to the formula, FAN_RPM_FACTOR should be 3840000. I've amended it.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: looks like it's still off, lemme change it to spaces instead of tab on next push.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: yup, I corrected fan_data array but missed out changes groups array. Corrected.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Added back in DT binding document.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Make sense, I've changed the debug message per your suggestion.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Changed to return -ENODEV.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++			      const char *buf, size_t count)
++{
++	struct i2c_client *client = to_i2c_client(dev);
++	int fan_idx = to_sensor_dev_attr(da)->index;
++	long new_value;
++	int status;
++
++	status = kstrtol(buf, 10, &new_value);
++	if (status < 0)
++		return -EINVAL;
```
karhin1990: nit: `return status;`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	const u8 hi_reg = SEL_FAN(fan, REG_TACH_TARGET_HIGH);
++	u8 high_byte = (new_target & 0x1fe0) >> 5;
++	u8 low_byte = (new_target & 0x001f) << 3;
++
++	i2c_smbus_write_byte_data(client, lo_reg, low_byte);
++	i2c_smbus_write_byte_data(client, hi_reg, high_byte);
++}
++
++static void read_fan_config_from_i2c(struct i2c_client *client, int fan_idx)
++
```
karhin1990, line 251: nit: extra line space

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	read_fan_from_i2c(client, &data->fan[fan_idx].target,
++			  SEL_FAN(fan_idx, REG_TACH_TARGET_HIGH),
++			  SEL_FAN(fan_idx, REG_TACH_TARGET_LOW));
++	read_fan_from_i2c(client, &data->fan[fan_idx].tach,
++			  SEL_FAN(fan_idx, REG_TACH_READ_HIGH),
++			  SEL_FAN(fan_idx, REG_TACH_READ_LOW));
++}
++
++static struct emc2303_fan_data *
++emc2303_update_fan(struct i2c_client *client, int fan_idx)
```
karhin1990: nit: `emc2303_update_fan_data`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	if (read_u8_from_i2c(client, SEL_FAN(fan_idx, REG_FAN_CONFIGURATION_1),
++			     &conf1) < 0)
++		return;
++
++	data->fan[fan_idx].rpm_control = (conf1 & 0x80) != 0;
++	data->fan[fan_idx].multiplier = 1 << ((conf1 & 0x60) >> 5);
++	data->fan[fan_idx].poles = ((conf1 & 0x18) >> 3) + 1;
++}
++
++static void read_fan_data(struct i2c_client *client, int fan_idx)
```
karhin1990: nit: `update_tach_target_and_reading`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	const u8 lo_reg = SEL_FAN(fan, REG_TACH_TARGET_LOW);
++	const u8 hi_reg = SEL_FAN(fan, REG_TACH_TARGET_HIGH);
++	u8 high_byte = (new_target & 0x1fe0) >> 5;
++	u8 low_byte = (new_target & 0x001f) << 3;
++
++	i2c_smbus_write_byte_data(client, lo_reg, low_byte);
++	i2c_smbus_write_byte_data(client, hi_reg, high_byte);
++}
++
++static void read_fan_config_from_i2c(struct i2c_client *client, int fan_idx)
```
karhin1990: nit: `update_fan_config`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	struct i2c_client *client = to_i2c_client(dev);
++	int fan_idx = to_sensor_dev_attr(da)->index;
++	long new_value;
++	int status;
++
++	status = kstrtol(buf, 10, &new_value);
++	if (status < 0)
++		return -EINVAL;
++	status = emc2303_set_pwm_enable(client, fan_idx, new_value);
++	return count;
```
karhin1990, line 488: Should return `status` if it is <0, else only return `count`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++		fan_data->valid = true;
++		fan_data->last_updated = jiffies;
++	}
++
++	mutex_unlock(&data->update_lock);
++	return fan_data;
++}
++
++static struct emc2303_fan_data *
++emc2303_update_device_fan(struct device *dev, struct device_attribute *da)
```
karhin1990: This wrapper function's name is quite confusing. For the sake of code readability, I would recommend to omit it and do line 293, 294 and 296 directly at its caller side.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	if (time_after(jiffies, fan_data->last_updated + HZ + HZ / 2)
++	    || !fan_data->valid) {
++		read_fan_config_from_i2c(client, fan_idx);
++		read_fan_data(client, fan_idx);
++		fan_data->valid = true;
++		fan_data->last_updated = jiffies;
++	}
++
++	mutex_unlock(&data->update_lock);
++	return fan_data;
```
karhin1990: Currently all the `attr_store` and `attr_show` functions will not return error even the driver fail to update fan data, because this function never return status. In this case, it will mislead the user because user thought the operation is succeeded even it's failing under the hood.

IMO, we should make this function return status, so that `attr_store` and `attr_show` functions have a chance to check if anything went wrong and return the right error code.

Whether or not we still want this function to return `fan_data` (via function param), I'm fine either as caller should have a way to dig out the fan_data from the `client` or `dev` object. So I leave it to you to make the call.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++};
++
++#define SEL_FAN(fan, reg) (reg + fan * 0x10)
++
++/*
++ * According to ME, EDGX[1:0] = 0b01; USECK = 0b0;
++ *
++ * TACH edges         = 5
++ * No. of poles       = 2
++ * Multiplier         = 1
```
karhin1990: Looking at how this macro being used, seems like it doesn't cater for multiplier yet (we will still multiply this value with the multiplier). If this is the case, I think you want to omit this comment.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++				    u16 new_target)
++{
++	const u8 lo_reg = SEL_FAN(fan, REG_TACH_TARGET_LOW);
++	const u8 hi_reg = SEL_FAN(fan, REG_TACH_TARGET_HIGH);
++	u8 high_byte = (new_target & 0x1fe0) >> 5;
++	u8 low_byte = (new_target & 0x001f) << 3;
++
++	i2c_smbus_write_byte_data(client, lo_reg, low_byte);
++	i2c_smbus_write_byte_data(client, hi_reg, high_byte);
++}
```
karhin1990: Need to return status. `i2c_smbus_write_byte_data` will return negative errno if the operation failed

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	mutex_lock(&data->update_lock);
++
++	if (rpm_target == 0)
++		fan->target = 0x1fff;
++	else
++		fan->target = clamp_val(
++			FAN_RPM_FACTOR * fan->multiplier / rpm_target,
++			0, 0x1fff);
++
++	write_fan_target_to_i2c(client, fan_idx, fan->target);
```
karhin1990: Check the status returned from this function and return the status if status < 0 
(CAUTIOUS: remember to unlock the mutex before return!)

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++
++	status = kstrtol(buf, 10, &new_value);
++	if (status < 0)
++		return -EINVAL;
++	status = emc2303_set_pwm_enable(client, fan_idx, new_value);
++	return count;
++}
++
++static SENSOR_DEVICE_ATTR_RO(fan1_input, fan_input, 0);
++static SENSOR_DEVICE_ATTR_RW(fan1_div, fan_div, 0);
```
karhin1990: Let's check with ME and see if there is any use case that dynamically change the clock div (e.g. ME want to change it during V&V, customer want to change it). If no, I will recommend not to support it.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++		goto exit_invalidate;
++	}
++
++	fan->multiplier = 8 / new_div;
++
++	/* update fan target if high byte is not disabled */
++	if ((fan->target & 0x1fe0) != 0x1fe0) {
++		u16 new_target = (fan->target * old_div) / new_div;
++		fan->target = min_t(u16, new_target, 0x1fff);
++		write_fan_target_to_i2c(client, fan_idx, fan->target);
```
karhin1990: Ditto: need to check the return status

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	}
++
++exit_invalidate:
++	/* invalidate fan data to force re-read from hardware */
++	fan->valid = false;
```
karhin1990: Looks like we need to do this in fan_target_store and pwm_enable_store as well, because these 2 functions did modify the fan registers (which may cause fan_data to be outdated) as well

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++/*
++ * set/ config functions
++ */
++
++/*
++ * Note: we also update the fan target here, because its value is
++ * determined in part by the fan clock divider.  This follows the principle
++ * of least surprise; the user doesn't expect the fan target to change just
++ * because the divider changed.
++ */
```
karhin1990: `the user doesn't expect the fan target to change just because the divider changed`: I'm a bit confused by this sentence, if user doesn't expect fan target to change, why we update fan target here?

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++		rpm = (FAN_RPM_FACTOR * fan->multiplier) / fan->tach;
++	return sprintf(buf, "%d\n", rpm);
++}
++
++static ssize_t
++fan_fault_show(struct device *dev, struct device_attribute *da, char *buf)
++{
++	struct emc2303_fan_data *fan = emc2303_update_device_fan(dev, da);
++
++	return sprintf(buf, "%d\n", (fan->tach & 0x1fe0) == 0x1fe0);
```
karhin1990: I'm not really sure if this is the right approach to check for fan fault, because the datasheet didn't says that top byte of TACH_READ reg is 0xff means fan is fault. 

If top byte of TACH_READ reg is 0xff means fan is fault - then in `fan_input_show` function, we should return rpm = 0 if `(fan->tach & 0x1fe0) == 0x1fe0`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,694 @@
...
++	if (status < 0)
++		return -EINVAL;
++	status = emc2303_set_pwm_enable(client, fan_idx, new_value);
++	return count;
++}
++
++static SENSOR_DEVICE_ATTR_RO(fan1_input, fan_input, 0);
++static SENSOR_DEVICE_ATTR_RW(fan1_div, fan_div, 0);
++static SENSOR_DEVICE_ATTR_RW(fan1_target, fan_target, 0);
++static SENSOR_DEVICE_ATTR_RO(fan1_fault, fan_fault, 0);
```
karhin1990: Similarly, let's understand if there is any use case that we need to read this attr. If there's no use case, don't support it.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
karhin1990: Learnt something new!

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: About the updating every 1.5s matter, I can't find a solid reason from many other examples. Referring to:
http://blog.foool.net/wp-content/uploads/linuxdocs/hwmon.pdf

Most just say:
"The XXXXX only updates its values each YY seconds; reading it more often will
do no harm, but will return ‘old’values."

So I'm guessing it is more for resource management to only update if time has passed 1.5s. I'll change valid to false as recommended based on our offline discussion.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: All subfunctions are changed to chain status to the top caller properly.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed extra line.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: corrected to the return statements.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Changed subfunction name to emc2303_update_fan_data.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: changed name to update_tach_target_and_reading.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: changed name to update_fan_config.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed this intermediate wrapper function and added the client retrieval codes into the caller themselves.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Re-wrote the status handling for all store and show functions.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Kar Hin you're right, the FACTOR equation do not cater for multiplier, hence I removed that line of comment already.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Changed to make sure all i2c_smbus_write_byte_data calls will check and return status.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Similarly, changed so that if status <0, code will goto the part the unlocks mutex and return status.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Checked with ME and DHW, changing divider dynamically will not be supported. Hence, I removed all fan1_div functions entirely. Initial div setting (according to HW regMap multiplier is 4, RNGX[1:0] = 0b10) shall be implemented in initialization (next PR).

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Added status handling.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: change fan->valid to false for all store functions (fan_target and pwm_enable).

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed the entire fan_div code, including this comment since it is not a required support.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Similarly to fan_div, fan_fault is not required by HW hence I removed the whole thing. Fault can be checked via software at higher level by reading and comparing fan_target and fan_input.

Extended comment:
Based on my understanding of the data sheet, 0xff on TACH means fan is not spinning (0 RPM). The original code assuming 0 RPM = faulty without comparing the fan_target is kinda funny because maybe the fan really is just not started instead of being actually broken.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed fan_div and pwm_enable after discussing with HW.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: It looks fine in text editor but somehow it is screwed up in the preview here. Can I get a pass 😂😂.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
+diff --git a/Documentation/devicetree/bindings/hwmon/emc2303.txt b/Documentation/devicetree/bindings/hwmon/emc2303.txt
+new file mode 100644
+index 00000000..a9eb7b1
+--- /dev/null
++++ b/Documentation/devicetree/bindings/hwmon/emc2303.txt
+@@ -0,0 +1,33 @@
++EMC2303 (I2C)
++
++This device is a RPM-based PWM Fan Speed Controller for 3 fans.
++Each fan can beconfigured individually:
```
karhin1990: nit: be configured

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++
++- The PWM mode:
++   0: PWM is disabled
++   3: RPM based PWM
++
++- The fan divisor (for RPM mesaurement)
```
karhin1990: Since we didn't let user to choose the fan divisor, I think you can omit this

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
karhin1990: Looks like you are removing `fault` instead of pwm_enable, right?

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++
++	/* invalidate fan data to force re-read from hardware */
++	fan->valid = false;
++
++	mutex_unlock(&data->update_lock);
++	return status;
++}
++
++static int
++emc2303_set_pwm_enable(struct i2c_client *client, int fan_idx, long enable)
```
karhin1990, line 353: Just curious, will the fan start to spin by itself once power is supplied to it? Or it will only start to spin when `emc2303_set_pwm_enable` is called? 

(You might not have the answer now since you don't have the hardware with you, but wanted to point this out so that you remember to sort it out to understand the fan chip behavior, i.e. what is the effect when setting pwmEnable to `1` or `0`. If setting pwmEnable to `1` simply selects whether to use close loop algorithm WITHOUT starting/stopping the fan, then we can hardcode it (set bit 7 of REG_FAN_CONFIGURATION_1) in the init function directly)

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++ * Factor by equations 4-2 and 4-3 from data sheet is 3840000
++ */
++#define FAN_RPM_FACTOR 3840000
++
++struct emc2303_fan_data {
++	bool			enabled;
++	bool			rpm_control;
++	bool			valid;		/* registers are valid */
++	u8				poles;
++	u8				multiplier;
```
karhin1990: Since we are going to hardcode the multiplier based on ME's recommendation and don't allow user to change the fan div, I think this `multiplier` element is no longer needed. 

In any where of the driver code that need the multiplier value, we should simply used the hardcoded value. Or, you can simply incorporate the hardcoded multiplier value into the `FAN_RPM_FACTOR` constant

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++
++exit_unlock:
++	mutex_unlock(&data->update_lock);
++	return status;
++}
++
++/*
++ * set/ config functions
++ */
++
```
karhin1990: nit: additional line space

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++ *
++ * Factor by equations 4-2 and 4-3 from data sheet is 3840000
++ */
++#define FAN_RPM_FACTOR 3840000
++
++struct emc2303_fan_data {
++	bool			enabled;
++	bool			rpm_control;
++	bool			valid;		/* registers are valid */
++	u8				poles;
```
karhin1990: Doesn't seems like this element is needed.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -0,0 +1,640 @@
...
++ * TACH edges         = 5
++ * No. of poles       = 2
++ * Internal Clk freq. = 32 kHz
++ *
++ * Factor by equations 4-2 and 4-3 from data sheet is 3840000
++ */
++#define FAN_RPM_FACTOR 3840000
++
++struct emc2303_fan_data {
++	bool			enabled;
```
karhin1990: Doesn't seems like this element is needed.

karhin1990 (commented):
Good work, we are almost there!

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
karhin1990: I means, removing fault is the right move. 

Just want to check if it's a typo on your reply (`Removed fan_div and pwm_enable after discussing with HW`), or your intention is to remove pwm_enable

NIwkoe: oh dear, ok I didn't realize you were commenting my comment. You're right I removed fan_fault. Typo in my reply.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: I think this is a very nice thought question. Here's my understanding:

What pwm_enable does is it will only toggle bit 7 to either use.
1. Direct setting Mode
2. Fan Speed Control Mode (PWM mode)

HW in fact wanted bit 7 of FAN CONFIG 1 to be set at default for RPM control according to the register map provided.

Based on Section 4.8 in the datasheet, when Direct Setting Mode is selected, Fan will spin up when Fan Drive Setting is set to anything greater than 00h. In this third party driver, original implementer did not link any function that changes this register and I doubt we will be using this mode. Also in Section 4.1, it is mentioned that "Whenever the Direct Setting mode is enabled, the current drive will be changed to what was last written into the Fan Driver Setting Register". Since "what was last written" is 0, my inference is it will just stop if we tamper with pwm_enable when fan is spinning.

Back to RPM mode, description in Section 4.8 mentioned when PWM mode is used, fan will spin up when TACH Target (aka fan_target) is changed (guarded by Valid TACH). So the answer to the original question is, until fan_target is written, fan will not spin.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed multiplier and implemented a more robust FAN_RPM_FACTOR macro.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed extra line and the random typo "/".

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed poles and implemented a more robust FAN_RPM_FACTOR macro.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed enabled and implemented a more robust FAN_RPM_FACTOR macro.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Fixed typo.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Removed fan divisor description.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -173,22 +170,24 @@ index 00000000..d65f2dd
 +#define SEL_FAN(fan, reg) (reg + fan * 0x10)
 +
 +/*
-+ * According to ME, EDGX[1:0] = 0b01; USECK = 0b0;
++ * Bloew is the fan configuration used in SmartRacks:
```
karhin1990: typo: `Below`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
karhin1990: Thanks for explaining, Wilson. 

> Back to RPM mode, description in Section 4.8 mentioned when PWM mode is used, fan will spin up when TACH Target (aka fan_target) is changed (guarded by Valid TACH). So the answer to the original question is, until fan_target is written, fan will not spin.

So it seems like rcu-service will never need to access the pwm_enable sysfs attribute, after the fan driver init has completed. In this case, seems like there is no need to expose `pwn_enable` attr to user at all (hence no need pwm_enable_store and pwm_enable_show function). 

Anyways, I'm okay to leave your driver code as it is.

karhin1990 (approved):

PR #47, opened 2022-01-24
Add initial register configs for Rev B Fan Panel and use the right way to retrieve i2c_client pointer

NIwkoe:
Inject initialization code to pre-load config register defaults requested by HW engineers.

A new subfunction set_init_config is added to set the initial values via i2c to the registers when the driver is being instantiated by tFanPanel via new_device.

Signed-off-by: wkoe <wilson.koe@ni.com>

### DIFF
--- a/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
+++ b/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
@@ -1,4 +1,4 @@
-From b250cba009721e7ed50e50aedfbfb44dfc59ca93 Mon Sep 17 00:00:00 2001
+From cd6eb15bfabd6b94191c70f0e02d5e403d3ff679 Mon Sep 17 00:00:00 2001
 From: wkoe <wilson.koe@ni.com>
 Date: Wed, 22 Dec 2021 23:17:08 -0800
 Subject: [PATCH] hwmon: add support for SMSC EMC2303
@@ -8,8 +8,8 @@ Signed-off-by: wkoe <wilson.koe@ni.com>
  .../devicetree/bindings/hwmon/emc2303.txt          |  30 +
  drivers/hwmon/Kconfig                              |  10 +
  drivers/hwmon/Makefile                             |   1 +
- drivers/hwmon/emc2303.c                            | 735 +++++++++++++++++++++
- 4 files changed, 776 insertions(+)
+ drivers/hwmon/emc2303.c                            | 711 +++++++++++++++++++++
+ 4 files changed, 752 insertions(+)
  create mode 100644 Documentation/devicetree/bindings/hwmon/emc2303.txt
  create mode 100644 drivers/hwmon/emc2303.c
 
@@ -84,10 +84,10 @@ index b033e67..611c9c8 100644
  obj-$(CONFIG_SENSORS_F71882FG)	+= f71882fg.o
 diff --git a/drivers/hwmon/emc2303.c b/drivers/hwmon/emc2303.c
 new file mode 100644
-index 00000000..b570810
+index 00000000..b2ae311
 --- /dev/null
 +++ b/drivers/hwmon/emc2303.c
-@@ -0,0 +1,735 @@
+@@ -0,0 +1,711 @@
 +/*
 + * emc2303.c - hwmon driver for SMSC EMC2303 fan controller
 + *
@@ -127,7 +127,7 @@ index 00000000..b570810
 + * 1        DRECK        1 = The CLK pin acts as a clock output and is a push-pull driver.
 + * 0        USECK        0 = Use internal oscillator for all tachometer measurements.
 + */
-+#define REG_CONFIGURATION_VALUE 			0x42
++#define REG_CONFIGURATION_VALUE   0x42
 +
 +/* Fan Configuration Registers at 32H, 42H and 52H (Section 6.14)
 + * bit      Field        Value
@@ -136,7 +136,7 @@ index 00000000..b570810
 + * 4:3      EDGx[1:0]     01 = 5 edges (2 poles), effective TACH multiplier is 1.
 + * 2:0      UDTx[2:0]    011 = 400 ms update interval.
 + */
-+#define REG_FAN_CONFIGURATION_1_VALUE		0xcb
++#define REG_FAN_CONFIGURATION_1_VALUE   0xcb
 +
 +/* Fan Configuration 2 Registers at 33H, 43H and 53H (Section 6.15)
 + * bit      Field        Value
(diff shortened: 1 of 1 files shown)


recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -581,6 +770,10 @@ index 00000000..9041766
 +	dev_info(&client->dev, "pwm fan controller: '%s'\n",
 +		 client->name);
 +
++	status = set_init_config(client);
++	if (status < 0)
++		return status;
```
karhin1990: It think it's better to do this before calling `devm_hwmon_device_register_with_groups`. 

Also, let's log some message if the status is < 0: `dev_err(&client->dev, "Failure when setting initial configuration. Status: %d\n", status);`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -110,6 +110,107 @@ index 00000000..9041766
...
++/* Configuration Register at 20H (Section 6.2)
++ * bit      Field        Value
++ * 7        MASK         0 = ALERT pin is unmasked.
++ * 6        DIS_TO       1 = The SMBus Time-Out function is disabled.
++ * 5        WD_EN        0 = Watchdog Timer does not operate continuously.
++ * 4:2      -            -
++ * 1        DRECK        1 = The CLK pin acts as a clock output and is a push-pull driver.
++ * 0        USECK        0 = Use internal oscillator for all tachometer measurements.
++ */
++#define REG_CONFIGURATION_VALUE 			0x42
```
karhin1990: Let use 3 spacing or 1 tab between the macro and the value: `#define REG_CONFIGURATION_VALUE   0x42`

Same for the macros defined below

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -110,6 +110,107 @@ index 00000000..9041766
...
++ * High Byte at 3BH, 4BH and 5BH
++ * bit      Field        Value
++ * 7:0      FxDF[12:5]   0000 1000
++ *
++ * Low Byte at 3AH, 4AH and 5AH
++ * bit      Field        Value
++ * 7:3      FxDF[4:0]    0 0000
++ * 2:0      -            -
++ *
++ * Drive Fail Band Registers as a whole = 256 (in decimal).
```
karhin1990: nit: `Number of tach counts used by the Fan Drive Fail detection circuitry = 256 (in decimal).`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -488,6 +589,94 @@ index 00000000..9041766
...
++		status = i2c_smbus_write_byte_data(client,
++					SEL_FAN(idx, REG_TACH_TARGET_LOW),
++					SEL_FAN(idx, REG_TACH_TARGET_LOW_VALUE));
++		if (status < 0)
++			return status;
++
++		/* Set TACH Target High Byte Registers. */
++		status = i2c_smbus_write_byte_data(client,
++					SEL_FAN(idx, REG_TACH_TARGET_HIGH),
++					SEL_FAN(idx, REG_TACH_TARGET_HIGH_VALUE));
```
karhin1990: Why are we start spinning the fan? I thought we just doing some initial configuration?
Can you check with ME why are they doing so?

Ultimately, the initial fan speed (upon rack power up) should not be hardcoded, it should follow the RPM that has been set by the user before they power off the rack. [currently, we set the initial fan speed based on `RPMAbsoluteMin` defined in the config file. see tFanDomain.cpp]

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
```diff
@@ -488,6 +589,94 @@ index 00000000..9041766
...
++	if (status < 0)
++		return status;
++
++	/* Iterate through all 3 fans to set the fan registers. */
++	for (idx = 0; idx < 3; idx++)
++	{
++		/* Set Fan Configuration Registers. */
++		status = i2c_smbus_write_byte_data(client,
++					SEL_FAN(idx, REG_FAN_CONFIGURATION_1),
++					SEL_FAN(idx, REG_FAN_CONFIGURATION_1_VALUE));
```
karhin1990: Why need to do `SEL_FAN` instead of `REG_FAN_CONFIGURATION_1_VALUE` directly?

Same to all settings below

karhin1990:
Just a common practice, start the first character of the commit message/PR title in capital letter:
`Add initial register configs for Rev B Fan Panel`

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Ok, thanks for the suggestion. I've moved the init to before registering hwmon device and also added the log message if there's error.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Standardized all MACROS to use 3 space in between macro statement and the value.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: Rephrased the sentence as per suggested.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: After checking with ME. We will only need to spin the fan after start button is pushed. These values were added during their experimentation phase and is indeed not needed for driver initialization. Having hardcoded value here will cause the fan to spin right when driver is installed (when RackDeviceManager is instantiated) BUT start button not pushed yet. Fan shall only start spin at RPMAbsoluteMin speed when Start Button is pushed. Hence, removing these TACH_TARGET init settings.

recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
NIwkoe: My mistake, this actually caused fan 2 and fan 3 to behave abnormally during my testing because all settings where added with 0x10 and 0x20 (silly mistake!). I have corrected them to only take the values directly without the SEL_FAN macro.

karhin1990 (approved):
Looks like there's typo in the title (i2client). Title suggestion:
`Add initial register configs for Rev B Fan Panel and use the right way to retrieve i2c_client pointer`

PR #49, opened 2022-02-28
RCU Rev B Porting Stage 1: Minimal base image

NIwkoe:
This is the first commit using the latest linux-toradex packages. Smartrack's Device tree is removed for the moment (and the eval board's is used instead) because it needs a major redo to support Rev B RCU. The goal is just to be able to build the image from the latest packages with meta-smartracks.

Some patches were amended in order for the build to succeed. They were:
(redo) add-support-for-diodes-pi4ioe5v96248.patch
(remove) alter-default-apalis-device-tree.patch

apalis-imx8.conf:
This conf is directly taken from toradex's original file. If we decided to use original names (that is, not creating a "-smartracks" custom MACHINE), we can just add our edits into this conf.

Note:
This commit onwards will break the build until smartracks-manifest is updated to the latest version.


buildconf/local.conf
```diff
@@ -22,7 +22,7 @@ ACCEPT_FSL_EULA = "1"
 # NOTE: SmartRack images are currently supported on Apalis-iMX8X only.
```
nwilsonni: update this note as well to iMX8

recipes-images/images/packagegroup-ni-cli.bb
```diff
@@ -18,7 +18,6 @@ RDEPENDS_packagegroup-ni-cli = "\
 SUMMARY_packagegroup-base-ni-cli = "Packages preinstalled into SmartRacks RCU image"
 RRECOMMENDS_packagegroup-base-ni-cli = "\
     vsftpd \
-    rauc \
```
HatsyRei, line 21: Might be a good idea to also omit udev-ni-rules from here. Looking at Toradex's rules for imx8, seems like the i2c numbering is different from what was used for imx8x in 99-ni.rules. Here's a comparison of 99-toradex.rules for imx8 and imx8x (right before support was dropped for imx8x):

imx8: https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/apalis-imx8/99-toradex.rules?h=dunfell-5.x.y

imx8x: https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/apalis-imx8x/99-toradex.rules?h=dunfell-5.x.y&id=d9b673df1b21c14dfb9db76a037698a0065a8ee7

I suppose we can drop in "udev-toradex-rules" in its place, so the default imx8 udev rules get installed instead. We can then build on top of it when we get to the DT part or something.

buildconf/local.conf
NIwkoe: Thanks for spotting this Nick, I've amended it.

recipes-images/images/packagegroup-ni-cli.bb
NIwkoe: Thanks for the suggestion Yan Huan, I've removed it for now. We shall add the new rules when we work on Rev B device tree.

conf/machine/apalis-imx8.conf
```diff
@@ -1,20 +1,20 @@
 #@TYPE: Machine
```
karhin1990: Only need this file if we are creating a custom MACHINE. Since we are re-using Toradex apalis-imx8 MACHINE, this file is no longer needed. All we need to do is to re-define the `KERNEL_DEVICETREE` param

recipes-images/images/packagegroup-ni-cli.bb
karhin1990: Looking at the latest rev, seems like the `99-ni.rules` files is still there.

karhin1990 (commented):
Good work, hope you find it interesting!

conf/machine/apalis-imx8.conf
karhin1990: I mean, you _could_ define `KERNEL_DEVICETREE` in the linux-toradex_5.4-2.3.x.bbappend to overwrite (via =) or append (via +=) the original dtb set in Toradex's apalis-imx8 machine .conf file

recipes-kernel/linux/linux-toradex_5.4-2.3.x.bbappend
```diff
@@ -6,14 +6,8 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
karhin1990, line 6: nit: Comment at line 3 and line 5 are outdated.

`Toradex defconfig`, `https://git.toradex.com/cgit/meta-toradex-nxp.git/tree/recipes-kernel/linux/linux-toradex_5.4-2.3.x.bb?h=dunfell-5.x.y`

recipes-images/images/packagegroup-ni-cli.bb
NIwkoe: I did not remove the file, I just excluded the bitbake recipe that adds 99-ni.rules since after we are done with device tree, we most likely need it again with imx8 changes. If it is cleaner to remove them altogether for this stae, I can do so too.

recipes-kernel/linux/linux-toradex_5.4-2.3.x.bbappend
NIwkoe: Updated outdated comments.

recipes-images/images/packagegroup-ni-cli.bb
karhin1990: Ah I see.. I'm okay to remain the file.

conf/machine/apalis-imx8.conf
NIwkoe: You're right Kar Hin, I've removed the .conf. Now that I think of it, it's kinda funny that I just duplicated the exact same machine conf 😁 into meta-smartracks without doing anything. When doing the next PR to add smartracks' custom device tree, I'll make sure to add the overwrite in linux-toradex_5.4-2.3.x.bbapppend. Right now since there's no custom device tree yet, let's remain what it is.

karhin1990: Sounds good!

karhin1990 (approved):

PR #50, opened 2022-03-15
Add imx8 smartrack into the rcu image

NIwkoe:
Continuing with the Rev B porting work, the next step is to add the device tree for imx8 and all its dependencies. This PR will include the following changes:
1. repatch u-boot
2. add imx8-apalis-smartracks.dts
3. add imx8-apalis-smartracks.dtsi (most changes here)
4. use KERNEL_DEVICETREE_append into linux-toradex_5.4-2.3.x.bbappend to add smartracks custom device tree

This PR is created because I was finally able to compile an image with the changes (just and only this reason for now) and wanted to create a checkpoint for device tree stuffs. Most nodes (io expander etc.) are copy paste from rev A. Some detailed changes (eg. renaming GPIO names etc.) are not in yet and will be worked in parallel with the review.

Testing Done:
No testing done yet. The image can be built and the imx8-apalis-smartracks.dtb device tree blob is successfully created.

*Please ignore the amount of commits :)


recipes-kernel/linux/linux-toradex_5.4-2.3.x.bbappend
```diff
@@ -6,8 +6,16 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
...
     file://0001-hwmon-add-support-for-SMSC-EMC2303.patch \
     "
 
 # Inject SmartRacks custom board device tree to kernel source device tree location
+do_configure_prepend() {
+    cp ${WORKDIR}/imx8-apalis-smartracks.dtsi ${WORKDIR}/git/arch/arm64/boot/dts/freescale
+    cp ${WORKDIR}/imx8-apalis-smartracks.dts ${WORKDIR}/git/arch/arm64/boot/dts/freescale
+}
+
+KERNEL_DEVICETREE_append = "freescale/imx8-apalis-smartracks.dtb"
```
karhin1990: Best practice of _append: Put a space before the content, i.e. `" freescale/imx8-apalis-smartracks.dtb"`

recipes-kernel/linux/files/imx8-apalis-smartracks.dts
```diff
@@ -0,0 +1,15 @@
+/dts-v1/;
+
+#include "imx8qp-apalis-v1.1.dtsi"
+#include "imx8-apalis-eval.dtsi"
```
karhin1990: I think it's better to create our board level dtsi from scratch rather than build on top the evaluation board, unless there are many commonalities between our evaluation board and the apalis eval board.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_i2c0_lvds1>;
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <100000>;
+};
+
+&iomuxc {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_rcugpio>, <&pinctrl_cam1_gpios>, <&pinctrl_dap1_gpios>,
```
HatsyRei: Seems like you've grouped all GPIO pads into pinctrl_rcugpio. I imagine we can remove the other pin groups (e.g. pinctrl_gpio3, pinctrl_gpio4, etc) from pinctrl-0? Are they still declared somewhere else?

karhin1990 (commented):
Have not finished reviewing. Figured it would be helpful to post my initial comments first as 1 of the comment might change the way you wrote the dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			fsl,pins = <
+				IMX8QM_MIPI_DSI1_I2C0_SCL_MIPI_DSI1_I2C0_SCL	0x04000020
+				IMX8QM_MIPI_DSI1_I2C0_SDA_MIPI_DSI1_I2C0_SDA	0x04000020
+			>;
+		};
+
+		/* Apalis BKL1_ON */
+		pinctrl_bkl_on: bkl-on {
+			fsl,pins = <
+				IMX8QM_LVDS0_GPIO00_LVDS0_GPIO0_IO00	0x00000021
```
karhin1990: Is this referring to Pin296? If yes, isn't it IMX8QM_LVDS0_GPIO00_LSIO_GPIO1_IO04?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			reg = <0 0x95000000 0 0x400000>;
+		};
+
+		vdevbuffer: vdevbuffer {
+			compatible = "shared-dma-pool";
+			reg = <0 0x90400000 0 0x100000>;
+			no-map;
+		};
+	};
+};
```
karhin1990: All these (except that 2 LVDS nodes where we need to enable them) were defined in imx8-apalis-v1.1.dtsi. We have included this dtsi, so I guess you can remove them from this file

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
+#include <dt-bindings/pwm/pwm.h>
+#include <dt-bindings/leds/leds-pca955x.h>
+
+/ {
```
karhin1990, line 4: Need to define a node for RTC, we will be using the external RTC as the default one
```
 aliases {
        rtc1 = &rtc_i2c;
        rtc0 = &rtc;
    };
```

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			interrupts = <29 IRQ_TYPE_LEVEL_LOW>;
+			micrel,led-mode = <0>;
+			reg = <7>;
+			reset-assert-us = <2>;
+			reset-deassert-us = <2>;
+			reset-gpios = <&lsio_gpio1 11 GPIO_ACTIVE_LOW>;
+			reset-names = "phy-reset";
+		};
+	};
+};
```
karhin1990: Similarly, this entire block were defined in imx8-apalis-v1.1.dtsi, which we have included. Unless you want to overwrite some of the default setting, else I guess you no need to re-define the same thing here again

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+/* Apalis CAN2 */
+&flexcan2 {
+	status = "disabled";
+};
+
+/* Apalis CAN3 (optional) */
+&flexcan3 {
+	status = "disabled";
+};
+
```
karhin1990, line 60: nit: Extra line space

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+/* Apalis HDMI1 */
+&hdmi {
+	compatible = "cdn,imx8qm-hdmi";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_hdmi_ctrl>;
+	firmware-name = "imx/hdmi/hdmitxfw.bin";
+	hdmi-ctrl-gpios = <&lsio_gpio1 30 GPIO_ACTIVE_HIGH>;
+	lane-mapping = <0x93>;
```
karhin1990: Not sure if we can get rid of these and simply leave line 238 (`status = "okay"`)

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+		pinctrl-names = "default";
+		pinctrl-0 = <&pinctrl_usb3503a>;
+		connect-gpios = <&lsio_gpio0 31 GPIO_ACTIVE_LOW>;
+		initial-mode = <1>;
+		intn-gpios = <&lsio_gpio1 1 GPIO_ACTIVE_LOW>;
+		refclk-frequency = <25000000>;
+		reg = <0x08>;
+		reset-gpios = <&lsio_gpio1 2 GPIO_ACTIVE_LOW>;
+	};
+};
```
karhin1990: Similarly, this entire block were defined in imx8-apalis-v1.1.dtsi, which we have included. Unless you want to overwrite some of the default setting, else I guess you no need to re-define the same thing here again

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+/* RCU's I2C1 */
+&i2c2 {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_lpi2c2>;
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <400000>;
+	/delete-node/rtc@68;
+};
```
karhin1990: Can we simply do the following?
```
&i2c2 {
    status = "okay";
    clock-frequency = <400000>;
};
```

Also, what is `/delete-node/rtc@68` for? Is it use to revert the setting in imx8-apalis-eval.dtsi?
If yes, then you don't need to do it if you don't include imx8-apalis-eval.dtsi (see my earlier comment related to this)

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+/* RCU's I2C2 */
+&i2c3 {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_lpi2c3>;
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <400000>;
+};
```
karhin1990: Can we simply do the following?
```
&i2c3 {
	status = "okay";
};
```
TMK, only RCU I2C bus 1 need to use 400KHz clock speed, the rest remain on 100KHz. You can confirm with EE again.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			>;
+		};
+
+		/* Apalis I2C3 (CAM) */
+		pinctrl_lpi2c3: lpi2c3grp {
+			fsl,pins = <
+				IMX8QM_SIM0_PD_DMA_I2C3_SCL				0x04000020
+				IMX8QM_SIM0_POWER_EN_DMA_I2C3_SDA		0x04000020
+			>;
+		};
```
karhin1990: Already defined in imx8-apalis-v1.1.dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+        compatible = "nxp,pca9674";
+        reg = <0x20>;
+        gpio-controller;
+        #gpio-cells = <2>;
+        label = "pca9674-5a830000.gpio";
+        /* IO0 and IO1 has to be output high by default */
+        lines-initial-states = <0x00>;
+        gpio-line-names =
+            "IOXPD2_IO0_0", "IOXPD2_IO0_1", "IOXPD2_IO0_2", "IOXPD2_IO0_3", "IOXPD2_IO0_4", "IOXPD2_IO0_5", "IOXPD2_IO0_6", "IOXPD2_IO0_7";
+    };
```
karhin1990, line 98: Text misalignment. Need to move more to the right

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+		/* Apalis I2C8 */
+		pinctrl_i2c0_lvds1: lvds1i2c0grp {
+			fsl,pins = <
+				IMX8QM_LVDS1_I2C0_SCL_LVDS1_I2C0_SCL	0x04000020
+				IMX8QM_LVDS1_I2C0_SDA_LVDS1_I2C0_SDA	0x04000020
+			>;
+		};
+
+		/* Apalis I2C10 */
```
karhin1990: Can we also specify this is RCU's I2C3? Just want to avoid confusion between RCU I2C and Apalis I2C

Same goes to other similar comments above

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+        /* For a pin to be configured as output, the bit within the bitmask should be set to 1.                                            */
+        /* Likewise, for a pin to be configured as input or output high, the bit within the bitmask should be set to 0.                    */
+        /* IO0_0 is LSB (bit at rightmost corner), and as we move left from LSB, we approach IO5_7 at the 48th bit.                        */
+        /* Everything between the 48th bit to MSB (bit at leftmost corner) does not represent anything (since IO expander is 48 bit only). */
+        
+        /* This property holds two consecutive 32-bit value to represent a 64-bit value. */
+        /* In binary, the hex value 0x00000000 00006000 is written as:                   */
+        /* 00000000 00000000 00000000 00000000 00000000 00000000 01100000 00000000       */
+        /* where bit 1s target IO1_5 and IO1_6 to set them as default output LOW         */
+        lines-initial-states = <0x00000000 0x00006000>;
```
karhin1990: Text misalignment

karhin1990: From [FCU Carrier Pinout spreadsheet](https://nio365.sharepoint.com/:x:/r/sites/SmartRack/_layouts/15/Doc.aspx?sourcedoc=%7B5851DCC2-D8D6-4933-84E4-E78E089BCA57%7D&file=FCU%20Carrier%20Pinout.xlsx&wdLOR=cD378E53B-9319-4BF0-9156-2EE226DF27D4&action=default&mobileredirect=true) seems like we want to set **IO1_4** and **IO1_5** to LOW instead

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+		pinctrl_rcugpio: rcugpiogrp {
+			fsl,pins = <
+				IMX8QM_FLEXCAN2_RX_LSIO_GPIO4_IO01			0x06000021 /* MXM3 11  */
+				IMX8QM_M40_I2C0_SCL_LSIO_GPIO0_IO06			0x06000021 /* MXM3 110 */
+				IMX8QM_UART1_RTS_B_LSIO_GPIO0_IO26			0x06000021 /* MXM3 114 */
+				IMX8QM_UART1_CTS_B_LSIO_GPIO0_IO27			0x06000021 /* MXM3 116 */
+				IMX8QM_FLEXCAN0_RX_LSIO_GPIO3_IO29			0x06000021 /* MXM3 12  */
+				IMX8QM_M40_I2C0_SDA_LSIO_GPIO0_IO07			0x06000021 /* MXM3 120 */
+				IMX8QM_M41_I2C0_SDA_LSIO_GPIO0_IO11			0x06000021 /* MXM3 122 */
+				IMX8QM_MIPI_DSI1_GPIO0_00_LSIO_GPIO1_IO22	0x06000021 /* MXM3 123 */
```
karhin1990: Text misalignment

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+				IMX8QM_USDHC1_DATA7_LSIO_GPIO5_IO22			0x06000021 /* MXM3 158 */
+				IMX8QM_FLEXCAN1_RX_LSIO_GPIO3_IO31			0x06000021 /* MXM3 16  */
+				IMX8QM_MLB_DATA_LSIO_GPIO3_IO28				0x06000021 /* MXM3 17  */
+				IMX8QM_ESAI0_TX0_LSIO_GPIO2_IO26			0x06000021 /* MXM3 177 */
+				IMX8QM_ESAI0_TX1_LSIO_GPIO2_IO27			0x06000021 /* MXM3 179 */
+				IMX8QM_FLEXCAN1_TX_LSIO_GPIO4_IO00			0x06000021 /* MXM3 18  */
+				IMX8QM_ESAI0_TX2_RX3_LSIO_GPIO2_IO28		0x06000021 /* MXM3 181 */
+				IMX8QM_ESAI0_TX3_RX2_LSIO_GPIO2_IO29		0x06000021 /* MXM3 183 */
+				IMX8QM_ESAI0_TX4_RX1_LSIO_GPIO2_IO30		0x06000021 /* MXM3 185 */
+				IMX8QM_ESAI0_TX5_RX0_LSIO_GPIO2_IO31		0x06000021 /* MXM3 187 */
```
karhin1990: Text misalignment

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+				IMX8QM_SPI0_CS0_LSIO_GPIO3_IO05				0x06000021 /* MXM3 227 */
+				IMX8QM_SPI2_SDI_LSIO_GPIO3_IO09				0x06000021 /* MXM3 229 */
+				IMX8QM_SPI2_SDO_LSIO_GPIO3_IO08				0x06000021 /* MXM3 231 */
+				IMX8QM_SPI2_CS0_LSIO_GPIO3_IO10				0x06000021 /* MXM3 233 */
+				IMX8QM_SPI2_SCK_LSIO_GPIO3_IO07				0x06000021 /* MXM3 235 */
+				IMX8QM_USB_SS3_TC2_LSIO_GPIO4_IO05			0x06000021 /* MXM3 262 */
+				IMX8QM_ESAI0_FSR_LSIO_GPIO2_IO22			0x06000021 /* MXM3 271 */
+				IMX8QM_ESAI0_FST_LSIO_GPIO2_IO23			0x06000021 /* MXM3 273 */
+				IMX8QM_USB_SS3_TC0_LSIO_GPIO4_IO03			0x06000021 /* MXM3 274 */
+				IMX8QM_SIM0_GPIO0_00_LSIO_GPIO0_IO05		0x06000021 /* MXM3 275 */
```
karhin1990: Text misalignment

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+				IMX8QM_UART0_RTS_B_LSIO_GPIO0_IO22			0x06000021 /* MXM3 6   */
+				IMX8QM_MLB_CLK_LSIO_GPIO3_IO27				0x06000021 /* MXM3 63  */
+				IMX8QM_M41_GPIO0_01_LSIO_GPIO0_IO13			0x06000021 /* MXM3 7   */
+				IMX8QM_UART0_CTS_B_LSIO_GPIO0_IO23			0x06000021 /* MXM3 8   */
+			>;
+		};
+
+		pinctrl_fec2: fec2grp {
+			fsl,pins = <
+				IMX8QM_ENET1_MDC_CONN_ENET1_MDC						0x06000020
```
karhin1990: For all pins listed under `iomuxc` section, can you also put a comment to list down its actual pin number (`/* MXM3 xyz   */`)

karhin1990: Text misalignment. There are many more misalignment in below

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			fsl,pins = <
+				IMX8QM_GPT1_CLK_DMA_I2C2_SCL			0x04000020
+				IMX8QM_GPT1_CAPTURE_DMA_I2C2_SDA		0x04000020
+			>;
+		};
+
+		/* Apalis I2C3 (CAM) */
+		pinctrl_lpi2c3: lpi2c3grp {
+			fsl,pins = <
+				IMX8QM_SIM0_PD_DMA_I2C3_SCL				0x04000020
```
karhin1990: Text misalignment

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+				IMX8QM_USDHC1_DATA2_CONN_USDHC1_DATA2		0x00000020
+				IMX8QM_USDHC1_DATA3_CONN_USDHC1_DATA3		0x00000020
+				IMX8QM_USDHC1_DATA4_CONN_USDHC1_DATA4		0x00000020
+				IMX8QM_USDHC1_DATA5_CONN_USDHC1_DATA5		0x00000020
+				IMX8QM_USDHC1_DATA6_CONN_USDHC1_DATA6		0x00000020
+				IMX8QM_USDHC1_DATA7_CONN_USDHC1_DATA7		0x00000020
+				/* On-module PMIC use */
+				IMX8QM_USDHC1_VSELECT_CONN_USDHC1_VSELECT	0x00000020
+			>;
+		};
```
karhin1990: Already defined in imx8-apalis-v1.1.dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+		/* Apalis SPDIF */
+		pinctrl_spdif0: spdif0grp {
+			fsl,pins = <
+				IMX8QM_SPDIF0_TX_AUD_SPDIF0_TX			0xc6000040
+				IMX8QM_SPDIF0_RX_AUD_SPDIF0_RX			0xc6000040
+			>;
+		};
```
karhin1990: Already defined in imx8-apalis-v1.1.dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+	pinctrl-names = "default", "state_100mhz", "state_200mhz";
+	pinctrl-0 = <&pinctrl_usdhc1>;
+	pinctrl-1 = <&pinctrl_usdhc1_100mhz>;
+	pinctrl-2 = <&pinctrl_usdhc1_200mhz>;
+	bus-width = <8>;
+	non-removable;
+	status = "okay";
+};
+
+/* Apalis MMC1 */
```
karhin1990: Is this the SD card slot for the RCU? Perhaps you can put some comment to explain what is this

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+&usdhc2 {
+	pinctrl-names = "default", "state_100mhz", "state_200mhz";
+	pinctrl-0 = <&pinctrl_usdhc2>, <&pinctrl_mmc1_cd>;
+	pinctrl-1 = <&pinctrl_usdhc2_100mhz>, <&pinctrl_mmc1_cd>;
+	pinctrl-2 = <&pinctrl_usdhc2_200mhz>, <&pinctrl_mmc1_cd>;
+	bus-width = <8>;
+	cd-gpios = <&lsio_gpio2 9 GPIO_ACTIVE_LOW>; /* Apalis MMC1_CD# */
+	no-1-8-v;
+	status = "okay";
+};
```
karhin1990: Can we simply do the following?
```
&usdhc2 {
	status = "okay";
};
```

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+/* On-module eMMC */
+&usdhc1 {
+	pinctrl-names = "default", "state_100mhz", "state_200mhz";
+	pinctrl-0 = <&pinctrl_usdhc1>;
+	pinctrl-1 = <&pinctrl_usdhc1_100mhz>;
+	pinctrl-2 = <&pinctrl_usdhc1_200mhz>;
+	bus-width = <8>;
+	non-removable;
+	status = "okay";
+};
```
karhin1990: Can we simply do the following?
```
&usdhc1 {
	status = "okay";
};
```

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+/* Apalis USBO1 */
+&usbotg1 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_usbotg1>;
+	adp-disable;
+	ci-disable-lpm;
+	hnp-disable;
+	over-current-active-low;
+	power-active-high;
+	srp-disable;
```
karhin1990: Can we get rid of these?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+		/* Apalis UART5 */
+		pinctrl_lpuart4: lpuart4grp {
+			fsl,pins = <
+				IMX8QM_M40_GPIO0_00_DMA_UART4_RX		0x06000020
+				IMX8QM_M40_GPIO0_01_DMA_UART4_TX		0x06000020
+			>;
+		};
+
+		/* Apalis USBH_EN */
```
karhin1990: From here onward till the end of `iomuxc` section, check the content with imx8-apalis-v1.1.dtsi. If the same thing already defined in imx8-apalis-v1.1.dtsi, you can exclude it from here, unless you want to overwrite some of the content

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+
+&jpegdec {
+	status = "okay";
+};
+
+&jpegenc {
+	status = "okay";
+};
+
+&ldb2_phy {
```
karhin1990, line 472: Can you put a comment on what this node is refer to? Same goes to other node that don't have comment

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+	pinctrl-0 = <&pinctrl_lpspi2>;
+	#address-cells = <1>;
+	#size-cells = <0>;
+	cs-gpios = <&lsio_gpio3 10 GPIO_ACTIVE_LOW>;
+
+	spidev1: spi@0 {
+		compatible = "toradex,evalspi";
+		reg = <0>;
+		spi-max-frequency = <4000000>;
+	};
```
karhin1990: Can we get rid of these?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+};
+
+/* RCU UART3 */
+&lpuart0 {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_lpuart0>;
+};
+
+/* RCU UART5 */
```
karhin1990: RCU only have 4 UART right? Might want to check with EE on which Apalis UART port is used for RCU UART 1, 2, 3, 4. I've emailed her just now.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+&lpuart4 {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_lpuart4>;
+	dma-names = "","";
+};
+
+&lsio_gpio0 {
```
karhin1990: What is the purpose of defining this node? Can you put a comment here? Same goes for other node below.

[DT source is quite hard to read and understand, so having adequate comment is very helpful for future references]

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+	gpio-line-names = "MXM3_279",
+			  "MXM3_277",
+			  "MXM3_135",
+			  "MXM3_203",
+			  "MXM3_201",
+			  "MXM3_275",
+			  "MXM3_110",
+			  "MXM3_120",
+			  "MXM3_1/GPIO1",
```
karhin1990: Why there is a `/GPIO1`?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+			cooling_maps_map0: map0 {
+				trip = <&pmic_alert0>;
+			};
+		};
+	};
+};
+
+&usb3phynop1 {
+	status = "okay";
+};
```
karhin1990: What is this for? Why are we enabling it?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+	srp-disable;
+};
+
+&usbphynop2 {
+	status = "okay";
+};
+
+&usbphy1 {
+	status = "okay";
+};
```
karhin1990: What is this for? Why are we enabling it?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -0,0 +1,1485 @@
...
+		};
+	};
+};
+
+&usb3phynop1 {
+	status = "okay";
+};
+
+/* Apalis USBH2, Apalis USBH3 and on-module Wi-Fi via on-module HSIC Hub */
+&usbh1 {
```
karhin1990, line 574: Need to explicitly enable this bus since we no longer include imx8-apalis-eval.dtsi

recipes-core/udev/files/99-ni.rules
```diff
@@ -1,19 +1,19 @@
...
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="i2c-[0-9]*-mux (chan_id 1)", SYMLINK+="I2C5_MUX1_CH1"
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="i2c-[0-9]*-mux (chan_id 2)", SYMLINK+="I2C5_MUX1_CH2"
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="i2c-[0-9]*-mux (chan_id 3)", SYMLINK+="I2C5_MUX1_CH3"
 ACTION=="add", ATTRS{iomem_base}=="0x5A070010", SYMLINK+="UART1"
-ACTION=="add", ATTRS{iomem_base}=="0x5A060010", SYMLINK+="UART2"
-ACTION=="add", ATTRS{iomem_base}=="0x5A080010", SYMLINK+="UART3"
-ACTION=="add", ATTRS{iomem_base}=="0x5A090010", SYMLINK+="UART4"
+ACTION=="add", ATTRS{iomem_base}=="0x5A090010", SYMLINK+="UART2"
+ACTION=="add", ATTRS{iomem_base}=="0x5A060010", SYMLINK+="UART3"
+ACTION=="add", ATTRS{iomem_base}=="0x5A0A0010", SYMLINK+="UART5"
```
karhin1990: We need to name the UART based on what has printed on the silkscreen. I think they only print UART 1, 2, 3 and 4. Need to check this out

nwilsonni: Silkscreen lists "AC PDU COMMUNICATION" and then numbers each port 1 - 6.  You can see silkscreen on 2022 WW09 Digital meeting notes

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
nwilsonni: That is correct - one of the user/ita I2C busses was to be 400kHz  I could not find where we had documented which one

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: This is pin286 from the pinout designer and it is to be muxed to LVDS0_GPIO0_IO00 for the BKL1_ON function.

NIwkoe: Ok sorry, I think I'm confused here too, the original mux is LSIO_GPIO1_IO04 for BKL1_ON, but the muxing provided by HW eng from pinout designer is LVDS0_GPIO0_IO00. Let me confirm again with them.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: I'll resolve all the text misalignments in my next push. My bad, it was due to me using tabs instead of spaces hence all the weird alignments.

karhin1990: Maybe you can change the tab setting in your editor (to a specific number of space), and then re-save the file. It might help you to update the tab on the entire file

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
karhin1990: Checked with Jing Wen, this should be RCU UART4

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Validated with EE, I2C1 and I2C2 on the RCU will be using 400kHz, the rest 100kHz.

karhin1990: Cool, I will mark this conversation as resolved now. Seems like EE should record this decision some where.

edit: will still keep this conversation open, since it also involve the comment to simplify `&i2c3` node

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: I'm not sure if we should remove them or not because these pin groups are still declared in imx8-apalis-v1.1.dtsi.

recipes-kernel/linux/linux-toradex_5.4-2.3.x.bbappend
NIwkoe: Thanks for sharing the best practice Kar Hin, I've added a space.

recipes-kernel/linux/files/imx8-apalis-smartracks.dts
NIwkoe: You're right, I've removed imx8-apalis-eval.dtsi from the include.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Verified with EE, the pin used for BKL1_ON for RCU will be **LVDS0_GPIO0_IO00**.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: The fact that there's so many duplication is because imx8-apalis-smartracks.dtsi is a copy-then-edit of imx8-apalis-v1.1.dtsi. I think my workflow was wrong, I only deleted the nodes that we didn't want, remained the ones we need and edited the additions accordingly. Now that I think of it, deleting in -smartracks won't make a difference if the node still exists in -v1.1. Anyway, I've remove all the duplicated nodes from -smartracks, thanks for pointing them out!

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Added RTC node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed duplicated node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed extra line.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified to only status = "okay".

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed duplicated node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified the i2c2 node.
You're right, it was due to -eval that I have to /delete-node/. Now that I removed -eval, this deletion is not needed and removed.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified 12c3 node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed duplicated node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: I've renamed all I2C and UART to follow RCU's index.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: By changing tabs to spaces, I believe all text misalignments are corrected.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Whoops, nice catch! changed to 0x00003000 for IO1_4 and IO1_5 LOW.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Ok, I've added /* MXM3 xyz */ comments for all muxed pins.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: For Apalis MMC1 (aka RCU's SD1), pins are redefined (reduce DATA lines from 8 bits to 4 bits) because SD only uses 4 data lines.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed duplicated node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Renamed to SD1 and added a brief explanatory comment.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified usdhc2 node, added also `bus-width = <4>; ` for SD card function.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified usdhc1 node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified usbotg1 node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Removed duplicated nodes.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: ldb2_phy is an LVDS signal, I've added a title comment above it.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Simplified lpspi2 node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Changed title comment to UART4.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: These where duplicated from -v1.1, I've removed them.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: These where duplicated from -v1.1, I've removed them.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Checked the pinout, this signal is needed for usbotg3. Since RCU do not use usbotg3, I've set this node to disabled.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: usbphynop2 is required by usbh1, hence setting it to okay, I've added an explanatory comment. Disabled usbphy1 since it is consumed by usbotg1, which is not used in RCU.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: You're right Kar Hin, I missed this, added status = "okay".

recipes-core/udev/files/99-ni.rules
NIwkoe: I've amended to UART4, aligned with EE's definition.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
karhin1990: Just to make sure we are on the same page. So we both agreed that Pin 286 will be used as the "back light ON" control pin for LVDS1. 

Looking at Apalis iMX8 datasheet, it seems to me Pin286 is referred as IMX8QM_**LVDS0_GPIO00**_**LSIO_GPIO1_IO04**:
image

If we are referring to the same pin, that means the value provided by the datasheet is not tally with the value provided by the HW Eng/pinout designer?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -2,203 +2,59 @@
...
-		};
-	};
+    aliases {
+        rtc0 = &rtc_i2c;
+        rtc1 = &rtc;
+    };
+
+    chosen {
+        stdout-path = &lpuart1;
+    };
```
karhin1990: Isn't this already specified in imx8-apalis-v1.1.dtsi?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -2,203 +2,59 @@
...
+        compatible = "panel-lvds";
+
+        status = "okay";
+
+        port {
+            panel2_lvds_in: endpoint {
+                remote-endpoint = <&lvds2_out>;
+            };
+        };
+    };
```
karhin1990: Not sure can we simply do:
```
/* Apalis LVDS1 */
&panel_lvds{
	status = "okay";
};

/* Apalis LVDS2 */
&panel2_lvds{
	status = "okay";
};
```

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -2,203 +2,59 @@
...
-	status = "disabled";
+    status = "disabled";
 };
 
 &adc1 {
-	status = "disabled";
+    status = "disabled";
 };
 
 &asrc0 {
```
karhin1990, line 32: Can you put a comment on what is this node refers to?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -213,111 +69,54 @@
...
-	status = "okay";
+    status = "okay";
 };
 
 &hsio_refa_clk {
-	status = "disabled";
+    status = "disabled";
 };
 
 &hsio_refb_clk {
```
karhin1990: Let's add some comment to tell what these 2 nodes refers to

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -451,1035 +250,429 @@
...
+                IMX8QM_LVDS1_GPIO00_LVDS1_PWM0_OUT      0x00000020 /* MXM3 239 */
+            >;
+        };
+
+        /* RCU's SD1_CD# */
+        pinctrl_mmc1_cd: mmc1cdgrp {
+            fsl,pins = <
+                IMX8QM_ESAI1_TX1_LSIO_GPIO2_IO09        0x00000021 /* MXM3 164 */
+            >;
+        };
```
karhin1990: These already been defined in imx8-apalis-v1.1.dtsi right?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -451,1035 +250,429 @@
...
+                IMX8QM_LVDS1_I2C0_SDA_LVDS1_I2C0_SDA    0x04000020 /* MXM3 283 */
+            >;
+        };
+
+        /* Apalis BKL1_ON */
+        pinctrl_bkl_on: bkl-on {
+            fsl,pins = <
+                IMX8QM_LVDS0_GPIO00_LVDS0_GPIO0_IO00    0x00000021 /* MXM3 286 */
+            >;
+        };
```
karhin1990, line 392: You can remove this as well if we end up agreed that Pin286 is referred as `IMX8QM_LVDS0_GPIO00_LSIO_GPIO1_IO04`, since this has already been defined in imx8-apalis-v1.1.dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -2,203 +2,59 @@
...
+        stdout-path = &lpuart1;
+    };
+
+    /* Apalis BKL1 */
+    backlight: backlight {
+        compatible = "pwm-backlight";
+        pinctrl-names = "default";
+        pinctrl-0 = <&pinctrl_bkl_on>;
+        enable-gpios = <&lsio_gpio1 4 GPIO_ACTIVE_HIGH>; /* Apalis BKL1_ON */
+    };
```
karhin1990: You can remove this as well if we end up agreed that Pin286 is referred as `IMX8QM_LVDS0_GPIO00_LSIO_GPIO1_IO04`, since this has already been defined in imx8-apalis-v1.1.dtsi

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
karhin1990: For which LVDS channel? We seems like using 2 LVDS channel

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -451,1035 +250,429 @@
...
-	interrupts = <GIC_SPI 474 IRQ_TYPE_LEVEL_HIGH>;
-	status = "okay";
-};
-
-&mu3_m0{
-	interrupts = <GIC_SPI 475 IRQ_TYPE_LEVEL_HIGH>;
-	status = "okay";
+    status = "okay";
+    pinctrl-names = "default";
+    pinctrl-0 = <&pinctrl_lpuart4>;
```
karhin1990, line 510: Doesn't seems like we include these 2 lines for RCU UART 1, 2 and 3. Shall we remove it for consistency?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -451,1035 +250,429 @@
...
 
 &rpmsg1{
-	status = "disabled";
+    status = "disabled";
 };
 
 &sai1 {
-	status = "disabled";
+    status = "disabled";
 };
```
karhin1990: Let's put a comment on each node to explain what are they refer to

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -451,1035 +250,429 @@
...
+    status = "okay";
 };
 
+/* This is the usbphy signal required by usbh1. */
 &usbphynop2 {
-	status = "okay";
+    status = "okay";
 };
 
 &usbphy1 {
```
karhin1990, line 585: Put a comment for this node

recipes-core/udev/files/99-ni.rules
```diff
@@ -1,19 +1,19 @@
 # Content in this file is derived from examples provided in udev-toradex-rules.bb in meta-toradex-bsp-common
 
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a810000.i2c", SYMLINK+="I2C1"
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="56246000.i2c", SYMLINK+="I2C2"
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a830000.i2c", SYMLINK+="I2C3"
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a820000.i2c", SYMLINK+="I2C4"
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="58226000.i2c", SYMLINK+="I2C5"
-ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="56226000.i2c", SYMLINK+="I2C6"
+ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a820000.i2c", SYMLINK+="I2C1"
```
karhin1990, line 3: Just curious (not a concern) - How do we know `5a820000` is refer to which physical I2C bus on the SOM? Is this documented in somewhere?

karhin1990 (commented):
Few more minor comments, then we are ready to go~

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: After syncing up with EE, we are more confused now. Base on datasheet, LSIO.GPIO1.IO04 makes sense, but according to EE's pin design, they are using the LVDS0.GPIO0.IO00 pin mux function. We will be posting a question to the toradex forum. Meanwhile, EE recommend to ignore this signal first for the time being.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: You're right Kar Hin, good catch, I've removed it.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Eventually, I removed the panel2 node altogether because after consulting EE, they only plan to use the 1st channel, second channel is just backup in case dual mode is used in the future. Hence, I simplified the panel_lvds node and removed the panel2_lvds node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Thanks for the recommendation, added more description to the node.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: I've removed these nodes as it is already disabled in imx8-apalis-v1.1.dtsi, don't need double definition.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Until we get a final answer from EE, I think I'll remain it as LVDS_GPIO0_IO00 first.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Until we get a final answer from EE, I think I'll remain it as LVDS_GPIO0_IO00 first, hence keeping this custom pinctrl-0 property for now.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: The reason only RCU UART 4 have these is because for RCU UART 1, 2 and 3, pinctrl-names and pinctrl-0 are already defined in imx8-apalis-v1.1.dtsi whereas RCU UART4 (aka lpuart4 node) is directly referencing its definition in imx8qm-ss-dma.dtsi.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Thanks for the reminder, I've added some comments for these nodes.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Thanks for the reminder, I've added some comments for these nodes.

recipes-core/udev/files/99-ni.rules
NIwkoe: I checked the definition of the i2c node. For example, 
```
	i2c2: i2c@5a820000 {
		reg = <0x5a820000 0x4000>;
		interrupts = <GIC_SPI 342 IRQ_TYPE_LEVEL_HIGH>;
		interrupt-parent = <&gic>;
		clocks = <&i2c2_lpcg 0>,
			 <&i2c2_lpcg 1>;
		clock-names = "per", "ipg";
		assigned-clocks = <&clk IMX_SC_R_I2C_2 IMX_SC_PM_CLK_PER>;
		assigned-clock-rates = <24000000>;
		power-domains = <&pd IMX_SC_R_I2C_2>;
		status = "disabled";
	};
```
definition from imx8-ss-dma.dtsi

And we know that RCU IC21 (from schematic)  corresponds to i2c2 node according to pinout design:
image

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
karhin1990: Ah I see!

recipes-core/udev/files/99-ni.rules
karhin1990: Got it!

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
karhin1990: Any update on this conversation? What is the final decision made?

karhin1990 (approved):

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: We will keep these pin groups since they were still declared in imx8-apalis-v1.1.dtsi.

HatsyRei: @karhin1990 Just had a discussion with Wilson on this, updating here to preserve context:

The other pin groups are still defined in imx8-apalis-v1.1.dtsi, and that is why they were not removed. I had some concerns with pinctrl_rcugpio containing pads already declared by other pin groups which are also assigned to iomuxc, e.g. [pinctrl_gpio3](https://git.toradex.com/cgit/linux-toradex.git/tree/arch/arm64/boot/dts/freescale/imx8-apalis-v1.1.dtsi?h=toradex_5.4-2.3.x-imx#n563), which would mean that the pad was assigned twice here. We can remove the individual pin groups from pinctrl-0, but seeing that the build passed without issues, it was deemed safe to do so. I'll be resolving this conversation for now.

karhin1990: Ah I see, thanks for the explanation. Hopefully everything goes well when we test it on the actual board (\*finger crossed\*)

PR #51, opened 2022-04-08
Fix io expander label and add udev-toradex-rules.bbappend

NIwkoe:
1. On the Rev B RCU, the mem address of I2C is different from that of Rev A. Hence, the label on the IO Expander should reflect the new mem address.
2. Because we use the original apalis-imx8 MACHINE instead of creating a custom "smartracks-imx8" MACHINE, bitbake bakes the udev-toradex-rules recipe. In Rev A, the 99-toradex.rules is an empty file because it takes from https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/99-toradex.rules?h=dunfell-5.x.y instead from the folder within the MACHINE. The 99-toradex.rules from apalis-imx8 defines a set of apalis specific SYMLINKs which are not applicable to smartracks (and makes the rules untidy). Hence, a bbappend is added to remove 99-toradex.rules entirely.

Checked that new image rootfs does not contain 99-toradex.rules.

### DIFF
--- a/recipes-core/udev/udev-toradex-rules.bbappend
+++ b/recipes-core/udev/udev-toradex-rules.bbappend
@@ -1,5 +1,9 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
 
+# The 99-toradex.rules from apalis-imx8 defines a set of apalis specific SYMLINKs which are not applicable to smartracks.
+# It is pulled in by bitbake because we use the default apalis-imx8 instead of creating a new custom smartracks MACHINE.
+# Hence, this bbappend is added to remove 99-toradex.rules entirely.
+
 do_install_append() {
     rm -f ${D}${sysconfdir}/udev/rules.d/99-toradex.rules
 }
--- a/recipes-core/udev/udev-toradex-rules.bbappend
+++ b/recipes-core/udev/udev-toradex-rules.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
+
+do_install_append() {
+    rm -f ${D}${sysconfdir}/udev/rules.d/99-toradex.rules
+}
--- a/recipes-core/udev/files/99-ni.rules
+++ b/recipes-core/udev/files/99-ni.rules
@@ -1,4 +1,5 @@
-# Content in this file is derived from examples provided in udev-toradex-rules.bb in meta-toradex-bsp-common
+# Content in this file is derived from examples provided in udev-toradex-rules.bb in meta-toradex-bsp-common, refer to
+# https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/apalis-imx8/99-toradex.rules?h=dunfell-5.x.y
 
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a820000.i2c", SYMLINK+="I2C1"
 ACTION=="add", KERNEL=="i2c-[0-9]*", ATTR{name}=="5a830000.i2c", SYMLINK+="I2C2"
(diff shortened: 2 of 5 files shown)


recipes-core/udev/udev-toradex-rules.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
```
HatsyRei, line 1: Since udev-toradex-rules is included by [packagegroup-tdx-cli.bb](http://git.toradex.com/cgit/meta-toradex-demos.git/tree/recipes-images/images/packagegroup-tdx-cli.bb?h=dunfell-5.x.y), I wonder if it is possible to create a packagegroup-tdx-cli.bbappend and just do

RRECOMMENDS_packagegroup-base-tdx-cli_remove = "udev-toradex-rules"

Could save the build process a few steps this way.

recipes-core/udev/udev-toradex-rules.bbappend
```diff
@@ -0,0 +1,5 @@
...
+
+do_install_append() {
+    rm -f ${D}${sysconfdir}/udev/rules.d/99-toradex.rules
+}
```
karhin1990, line 9: Can you put a comment on why we are doing this? I like your explanation in this PR, if we have that in the code then we won't lost the context!

recipes-core/udev/udev-toradex-rules.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files/:"
+
+do_install_append() {
+    rm -f ${D}${sysconfdir}/udev/rules.d/99-toradex.rules
```
karhin1990, line 8: I noticed we didn't port the last 5 lines from [99-toradex.rules](https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/apalis-imx8/99-toradex.rules?h=dunfell-5.x.y) to our own udev rule:

image

Do we need that?

NIwkoe: We will not need line 14 as that's specific for adc. mmcblk at 5b030000 (lines 15 and 16) is actually our sd card slot, according to Rev A's design, there were no SYMLINKs for the sd card so I assumed it is not needed (The only links used in rcu-service are the I2C and UART ones).
Following this idea, sounds like the watchdog symlink is not needed as well? it has a "apalis-" beginning
image
RCU service doesn't use this link and I'm not sure if it is consumed by any other non rcu-service code or not.
We can
1. Not include all these apalis specific links (means have to remove apalis-watchdog or rename to just "watchdog").
2. Include them in kernel as smartracks' placeholders by changing a different name (add sd card link but not with apalis label) for future use.

Any thoughts?

HatsyRei: We never made use of sd symlinks and watchdog, so it should be fine to exclude those.

karhin1990: I've also no idea on whether the symlink is needed by any scripts or program that are part of Apalis BSP. 

Perhaps we can remain them, and put a comment indicates that these rules were originated from toradex's udev rule (https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-core/udev/files/apalis-imx8/99-toradex.rules?h=dunfell-5.x.y).

recipes-core/udev/udev-toradex-rules.bbappend
NIwkoe: Thanks for the suggestion Kar Hin. I've added the comment into the bbappend recipe.

recipes-core/udev/udev-toradex-rules.bbappend
NIwkoe: Great! Since there's already an existing comment at the start of the file, I just added the link to the original .rules file.

karhin1990 (approved):

recipes-core/udev/udev-toradex-rules.bbappend
HatsyRei: cc: @karhin1990 @NIwkoe @NI-CHO 

Turns out there has been an issue with GitHub (or me using GitHub), that caused my reviews to stay in "pending" mode. These reviews were not visible to anyone but me. @NI-CHO alerted me to the fact that she did not see my comments on her PR! This is a comment I placed on Wilson's PR in April, but did not get submitted. 🤦‍♂️

karhin1990: What?!! 😮  How did you solve the issue then?

HatsyRei: I had to submit an empty review by going to the "Files changed" tab and click on "Review changes". Sigh. Should have suspected that something was weird when nobody responded to my review... :D

karhin1990: I see.. Glad that you found it out.

NIwkoe: Whoops, what's going on? Did I miss anything fun🤔

PR #52, opened 2022-04-12
Add i2c mux into device tree

NIwkoe:
In the RCU design, an i2c mux is connected to RCU's I2C5 to control several Temperature Sensors. The mux's selector is connected to two GPIOs from the RCU as control pins. This i2c mux node is ported from Rev A. Changes made were:
1. mux address label (follow its parent i2c port)
2. parent i2c node
3. mux gpios (gpio group and line number change)

### DIFF
--- a/recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
@@ -16,7 +16,7 @@
         status = "okay";
     };
 
-    /* I2C Analog Mux attached to Apalis I2C5 */
+    /* I2C Analog Mux attached to RCU's I2C5 */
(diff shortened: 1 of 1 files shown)


recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -15,6 +15,37 @@
     panel_lvds: panel-lvds {
         status = "okay";
     };
+
+    /* I2C Analog Mux attached to Apalis I2C5 */
```
karhin1990: nit: I guess you want to say `RCU I2C5`

karhin1990 (approved):

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: Oh dear yes, silly copy-paste error. Amended comment.

PR #53, opened 2022-05-10
Instantiate Onboard Temp sensor through device tree

NI-CHO:
Add the onboard temperature sensor to the device tree.
Tested on Gamora rack that the onboard temperature can be detected and driver can be loaded.

Signed-off-by: NI-CHO <cheng.hong.ong@ni.com>

### DIFF
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -4,6 +4,7 @@
     aliases {
         rtc1 = &rtc_i2c;
         rtc0 = &rtc;
+        temp0 = &temp_i2c;
     };
 
     reg_3v3: regulator-3v3 {
@@ -227,6 +228,13 @@
     rtc_i2c: rtc@68 {
         compatible = "dallas,ds3232";
         reg = <0x68>;
+    };
+    
+    /* Onboard Temperature Sensor TMP1075 */
+    temp_i2c: temp@4F {
+        status = "okay";
+        compatible = "lm75";
+        reg = <0x4F>;
     };
     
     /* 16-bit LED driver */
--- a/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
+++ b/recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
@@ -4,7 +4,6 @@
     aliases {
         rtc1 = &rtc_i2c;
         rtc0 = &rtc;
-        temp0 = &temp_i2c;
     };
 
     reg_3v3: regulator-3v3 {
@@ -232,9 +231,9 @@
     
     /* Onboard Temperature Sensor TMP1075 */
     temp_i2c: temp@4F {
-        status = "okay";
-        compatible = "lm75";
+        compatible = "national,lm75";
         reg = <0x4F>;
+        status = "okay";
     };
     
     /* 16-bit LED driver */


recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -4,6 +4,7 @@
     aliases {
         rtc1 = &rtc_i2c;
         rtc0 = &rtc;
+        temp0 = &temp_i2c;
```
karhin1990: Just for documentation purpose - Can you put a comment on why we are defining this alias? If you happen to see it from any online reference, putting the link here will be helpful as well ;)

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -229,6 +230,13 @@
         reg = <0x68>;
     };
     
+    /* Onboard Temperature Sensor TMP1075 */
+    temp_i2c: temp@4F {
+        status = "okay";
+        compatible = "lm75";
```
karhin1990: Isn't we have to choose 1 of the entry under this [list](https://git.toradex.com/cgit/linux-toradex.git/tree/Documentation/devicetree/bindings/hwmon/lm75.txt?h=toradex_5.4-2.3.x-imx#n4)? The closest I find is `"national,lm75"`

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
NI-CHO: No specific reason for defining this alias. I'm just adding onboard temperature by referring to the RTC device.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
NI-CHO: Oops. I'm not aware of this. I will modify it to "national,lm75"

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
```diff
@@ -229,6 +230,13 @@
         reg = <0x68>;
     };
     
+    /* Onboard Temperature Sensor TMP1075 */
+    temp_i2c: temp@4F {
+        status = "okay";
```
HatsyRei: Nit: Let's put status as the last property in the node, for consistency's sake.

recipes-kernel/linux/files/imx8qxp-apalis-smartracks.dtsi
karhin1990: Ah I see. It seems to me this alias is not needed, @HatsyRei do you know if it is necessary?

HatsyRei: @karhin1990 Hm, I don't think it is needed either. The rtc aliases are in there because the original device tree by Toradex had them, so we did not remove them.

I tried looking at what aliases are used for ([SO link](https://stackoverflow.com/questions/17738709/dts-file-explanation-aliases)), and as far as we are concerned, I don't think rcu-service makes use of the created aliases. It should be safe to omit this for the newly added onboard temp sensor.

cc: @NI-CHO

HatsyRei (approved):

karhin1990 (approved):
Nice!

PR #55, opened 2022-06-01
changes to support Rev B bringup

NIwkoe:
Add changes made during bring up:
1. Fix Device Tree not built into image
2. Disable all USB interface
3. Fix all GPIO muxing issues to enable all I/O buses
4. Enable Ethernet Port 2

### DIFF
--- a/buildconf/local.conf
+++ b/buildconf/local.conf
@@ -24,6 +24,9 @@ ACCEPT_FSL_EULA = "1"
 
 MACHINE ?= "apalis-imx8"
 
+# NOTE: Append KERNEL_DEVICETREE to insert imx8-apalis-smartracks.dtb into
+# apalis-imx8.conf. This will ensure the dtb to be built into our custom image.
+
 KERNEL_DEVICETREE_append = " freescale/imx8-apalis-smartracks.dtb"
 
 #
--- a/buildconf/local.conf
+++ b/buildconf/local.conf
@@ -24,6 +24,8 @@ ACCEPT_FSL_EULA = "1"
 
 MACHINE ?= "apalis-imx8"
 
+KERNEL_DEVICETREE_append = " freescale/imx8-apalis-smartracks.dtb"
+
 #
 # There are also a selection of emulated machines available which can boot and run
 # in the QEMU emulator:
--- a/recipes-bsp/u-boot/files/alter-default-apalis-device-tree.patch
+++ b/recipes-bsp/u-boot/files/alter-default-apalis-device-tree.patch
@@ -1,35 +1,26 @@
-From d38c215a0fbc4023ceaeb13bb873d5dd940a9e15 Mon Sep 17 00:00:00 2001
+From 98b9a3eafa5e664c2f5d6aa2a09f57c4c44fa989 Mon Sep 17 00:00:00 2001
 From: wkoe <wilson.koe@ni.com>
 Date: Mon, 14 Mar 2022 20:53:28 -0700
 Subject: [PATCH] alter default apalis device tree
 
 Signed-off-by: wkoe <wilson.koe@ni.com>
 ---
- include/configs/apalis-imx8.h | 4 +++-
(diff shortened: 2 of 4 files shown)


HatsyRei (approved):

recipes-bsp/u-boot/files/alter-default-apalis-device-tree.patch
```diff
@@ -1,35 +1,26 @@
...
+diff --git a/configs/apalis-imx8_defconfig b/configs/apalis-imx8_defconfig
+index 42700e4571..4b087e36dd 100644
+--- a/configs/apalis-imx8_defconfig
++++ b/configs/apalis-imx8_defconfig
+@@ -17,7 +17,7 @@ CONFIG_FIT=y
+ CONFIG_FIT_VERBOSE=y
+ CONFIG_OF_SYSTEM_SETUP=y
+ CONFIG_USE_PREBOOT=y
+-CONFIG_PREBOOT="setenv fdtfile ${soc}-apalis${variant}-${fdt_board}.dtb"
++CONFIG_PREBOOT="setenv fdtfile imx8-apalis-smartracks.dtb"
```
karhin1990, line 20: I think we should not patch the original defconfig from Toradex. Probably the right way to alter the kernel config (to add Smart Racks specific customization) is to do it via fragment file, like how we enable the kernel module for some sensor.

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
```diff
@@ -610,13 +668,12 @@
 
 /* Apalis USBH2, Apalis USBH3 and on-module Wi-Fi via on-module HSIC Hub */
 &usbh1 {
-    vbus-supply = <&reg_usb_host_vbus>;
-    status = "okay";
+    status = "disabled";
```
karhin1990, line 671: Why are we disabling USB? 

The PR from @HatsyRei (on rcu-service) seems to release GPIO pins that are consumed by USB_OTG, so I thought we are enabling USB_OTG (and USB 2.0) in DT for now. I think we should either:
- Release USB_OTG pins from rcu-service's GPIO manager and enable USB_OTG (as well as USB 2.0) in DT; or
- Remain the USB_OTG pins as GPIO in rcu-service, disable both USB_OTG and USB 2.0 in DT

Now we are in a mixing mode

buildconf/local.conf
```diff
@@ -24,6 +24,8 @@ ACCEPT_FSL_EULA = "1"
 
 MACHINE ?= "apalis-imx8"
 
+KERNEL_DEVICETREE_append = " freescale/imx8-apalis-smartracks.dtb"
```
karhin1990, line 30: nit: Can you put a comment here on why we put this line here instead of in `linux-toradex_5.4-2.3.x.bbappend`?

recipes-kernel/linux/files/imx8-apalis-smartracks.dtsi
NIwkoe: My bad for the confusion. I've clarified the requirements at @HatsyRei's rcu-service PR. Right now, we will switch off all onboard USB (OTG and hub) and release pins 262 and 274 for GPIO functions.

karhin1990: Sounds good!

recipes-bsp/u-boot/files/alter-default-apalis-device-tree.patch
NIwkoe: Synced up with @karhin1990 offline. Since this is u-boot's config instead of the linux kernel's config, we will not use fragment file and maintain this patching method.

buildconf/local.conf
NIwkoe: Added brief explanation of the addition of this append.

karhin1990 (approved):

PR #59, opened 2022-06-20
Install Mfg Config file

NI-CHO:
Place empty content mfgconfig.json file to the etc path
Signed-off-by: NI-CHO <cheng.hong.ong@ni.com>


recipes-core/rcu-config/files/mfgconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    ],
+    "FanPanels": [
+        {
+        }
+    ],
+    "FanDomains": [
+        {
+        }
+    ],
+    "PxiChassis": [
```
nwilsonni: PXI chassis is optional, so should not be in default file

recipes-images/images/packagegroup-ni-cli.bb
```diff
@@ -25,6 +25,7 @@ RRECOMMENDS_packagegroup-base-ni-cli = "\
     udev-ni-rules \
     keyutils \
     cifs-utils \
+    rcu-config \
```
karhin1990, line 28: nit: rcu-config**s**, since we have more than 1 config file

recipes-core/rcu-config/files/mfgconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
karhin1990: I doubt if it's worth to populate these placeholder content.. It is not a "default" content either, because the arrays are empty, so I _guess_ rcu-service will still fail to start if this file is used.

The downside I see for creating this placeholder content is - whenever we supports a new field, we have to come and update this file. Same comment applies to the userconfig.json

karhin1990 (approved):

recipes-core/rcu-config/files/mfgconfig.json
NI-CHO: If putting total blank content in the config file, it will halt the rcu-service. I think not a concern to mfgconfig.json since it is a required configuration in order for starting rcu-service. However, the blank userconfig will halt the rcu-service during std::make_unique<tUserJsonConfigManager>("/etc/userconfig.json")
_Example error: Failed Parsing /etc/userconfig.json_

I can think of 2 ways for not putting the placeholder content with fields:
1. blank content in the config file - requires handling it to continue when it is null at cJSONTools -  createJSONRootFromFile / create another createJSONRootFromFile for user config only
2. put below placeholder in the config file - require handling it to continue when it is not a JSON array through tUserJsonConfigManager::_findElementById
```
{
}
```
Do you see any better thoughts other than these?

karhin1990: Ah, I see what you means (but I don't agree that it will halt rcu-service, the most you will just receive an exception, which you can handle it with a try-catch block 😝 ).

From the options you provide, personally I feel the 2nd option looks cleaner.

kengsooncheah:
Just curious, why do we need the json file with empty content?

NI-CHO:
> Just curious, why do we need the json file with empty content?

The reason for putting blank content and installing them by default is without the need of checking if the config file exists in "rcu-service.cpp".
I will try to put a clearer comment in the git commit message - Install default blank config files because without the need of checking their existence in rcu-service

kengsooncheah (approved):

kengsooncheah (approved):

karhin1990 (approved):

karhin1990:
Let's update the PR title and description since we no longer install userconfig.json

PR #61, opened 2022-06-24
Enforce running rcu-hostname.service before avahi-daemon.service

HatsyRei:
On RCU boot, avahi-daemon may register default Apalis hostname to DNS server instead of the `ni-smartracks` hostname if avahi-daemon runs prior to rcu-hostname. Changes in this PR should allow rcu-hostname to always run before avahi-daemon.

### DIFF
--- a/recipes-core/rcu-hostname/files/rcu-hostname.service
+++ b/recipes-core/rcu-hostname/files/rcu-hostname.service
@@ -2,6 +2,7 @@
 #ConditionPathExists=|!/etc/hostname
 Wants=network-pre.target
 Before=network-pre.target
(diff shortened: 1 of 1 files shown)


karhin1990 (approved):
Good finding!

recipes-core/rcu-hostname/files/rcu-hostname.service
```diff
@@ -2,6 +2,7 @@
 #ConditionPathExists=|!/etc/hostname
 Wants=network-pre.target
 Before=network-pre.target
+Before=avahi-daemon.service
```
karhin1990, line 6: Let's add a comment here on why we are doing this

PR #62, opened 2022-06-27
Remove Conflicting Toradex Packages from RCU Image

HatsyRei:
Toradex's [packagegroup recipe](https://git.toradex.com/cgit/meta-toradex-demos.git/tree/recipes-images/images/packagegroup-tdx-cli.bb?h=dunfell-5.x.y#n44) installs `set-hostname` and `udev-toradex-rules`. The `set-hostname` package was found to race against `rcu-hostname` to set RCU hostname on first boot.

`udev-toradex-rules` was previously included, but deleted in https://github.com/ni/meta-smartracks/commit/1490ccec2d3e19fb0ff72f4c92c28c146f80bc99. This PR excludes it entirely from RCU image build.

Changes tested on a local build, rootfs tar inspected to confirm exclusion of both packages.

### DIFF
--- a/recipes-images/images/smartracks-minimal-image.bb
+++ b/recipes-images/images/smartracks-minimal-image.bb
@@ -32,6 +32,8 @@ IMAGE_LINGUAS = "en-us"
 
 CONMANPKGS ?= "connman connman-plugin-loopback connman-plugin-ethernet connman-client"
 
+BAD_RECOMMENDATIONS = "set-hostname udev-toradex-rules"
+
 IMAGE_INSTALL += " \
     packagegroup-boot \
     packagegroup-basic \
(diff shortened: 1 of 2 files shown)


recipes-images/images/smartracks-minimal-image.bb
```diff
@@ -32,6 +32,8 @@ IMAGE_LINGUAS = "en-us"
 
 CONMANPKGS ?= "connman connman-plugin-loopback connman-plugin-ethernet connman-client"
 
+BAD_RECOMMENDATIONS = "set-hostname udev-toradex-rules"
```
karhin1990, line 36: Let's add some comment to explain the intention of adding this line. I'm sure we will forget about the context after 3 month 😆

karhin1990 (approved):

recipes-images/images/smartracks-minimal-image.bb
HatsyRei: Setting a new record for forgetting to put comments  :D

PR #64, opened 2022-07-04
Support RCU uSD Port Rework to Apalis SD1

HatsyRei:
Unconfirmed changes, pending HW team confirmation.

It was recently discovered that the GPIO pins muxed from Apalis MMC1 (DT: usdhc2) were causing the remaining MMC1 pins in use to operate at higher voltage levels (2.1V ~ 2.2V) than expected (1.8V). A rework which routes uSD signals to Apalis SD1 (DT: usdhc3) has been dev tested and is shown to resolve these uSD voltage issues.


kengsooncheah:
I guess we'll need to shelve this until Rev C RCU (if the change is indeed needed)?

HatsyRei:
Closing since a different fix to address this issue has already been merged.

PR #69, opened 2022-08-15
Support RCU I2C3 Analog Mux

HatsyRei:
Analog I2C mux on RCU I2C3 may be present if 3-phase PEP is connected on a rack. This mux is not present on racks with single-phase PEPs, and the mux channels will not be enumerated by the driver if so.

Testing needs to be done once a 3P PEP capable rack is available.


karhin1990 (approved):
Looks good to me

kengsooncheah (approved):

PR #71, opened 2022-08-23
Disable Mender-client running as a system service.

lingchin9:
To solve the [bug about the mender error messages in Linux message log (/var/log/messages)](https://dev.azure.com/ni/DevCentral/_workitems/edit/2121915). Mender-client is enabled as a system service and periodically checking for updates from a dummy server in the **_meta-mender/meta-mender-core/recipes-mender/mender-client.inc_** file.

The _mender-client_%.bbappend_ file is added to disable the Mender-client running as a system service as refer to the [Customize Mender | Mender documentation](https://docs.mender.io/system-updates-yocto-project/customize-mender#disabling-mender-as-a-system-service).

### DIFF
--- a/recipes-core/mender/mender-client_%.bbappend
+++ b/recipes-core/mender/mender-client_%.bbappend
@@ -1,0 +1,1 @@
+SYSTEMD_AUTO_ENABLE = "disable"
--- a/recipes-core/mender/mender-client_%.bbappend
+++ b/recipes-core/mender/mender-client_%.bbappend
@@ -1,4 +1,4 @@
 # Disable Mender to run as a system service automatically at boot as refer to 
 # https://docs.mender.io/system-updates-yocto-project/customize-mender#disabling-mender-as-a-system-service
-# Uncomment below to enable mender running as a systemd service as it has been enabled in meta-mender/meta-mender-core/recipes-mender/mender-client.inc
+# Comment out line below to enable mender running as a systemd service as it has been enabled in meta-mender/meta-mender-core/recipes-mender/mender-client.inc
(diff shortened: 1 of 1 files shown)


recipes-core/mender/mender-client_%.bbappend
```diff
@@ -0,0 +1 @@
+SYSTEMD_AUTO_ENABLE = "disable"
```
karhin1990, line 4: Let's put some comments so that it's more readable. E.g.
```
# Disable the service that starts Mender at boot because we don't need it
# https://docs.mender.io/system-updates-yocto-project/customize-mender#disabling-mender-as-a-system-service
SYSTEMD_AUTO_ENABLE = "disable"
```

karhin1990 (approved):

kengsooncheah (approved):

HatsyRei (approved):

recipes-core/mender/mender-client_%.bbappend
```diff
@@ -0,0 +1,4 @@
+# Disable Mender to run as a system service automatically at boot as refer to 
+# https://docs.mender.io/system-updates-yocto-project/customize-mender#disabling-mender-as-a-system-service
+# Uncomment below to enable mender running as a systemd service as it has been enabled in meta-mender/meta-mender-core/recipes-mender/mender-client.inc
```
karhin1990: I guess you want to say "Comment out line below" instead of "Uncomment below"?

lingchin9: Oh ya, thank you for the reminder

PR #81, opened 2022-12-06
Customize RCU IP Settings via systemd-networkd Interface

HatsyRei:
PR to address [US 2195506](https://dev.azure.com/ni/DevCentral/_workitems/edit/2195506). Changes made to append eth0 and eth1 network configuration as part of systemd installation. After some consideration, these network configuration files will be installed in the persistent data partition (/data/systemd/network) - Symlinks to these files located in /data will be created where systemd-networkd expects them to be (/etc/systemd/network). 

Testing done on RCU + fanout board setup, eth0 works properly after changes were made.

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -1,0 +1,29 @@
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI += " \
+    file://eth0.network \
+    file://eth1.network \
+"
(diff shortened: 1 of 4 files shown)


recipes-core/systemd/systemd_%.bbappend
```diff
@@ -0,0 +1,29 @@
...
+"
+
+PACKAGECONFIG_append = " networkd"
+
+do_install_append() {
+
+    # We install actual eth0 and eth1 network configuration files into persistent data partition
+    # allowing these settings to persist across Mender updates
+    install -d ${D}/data/systemd/network/
+    install -d ${D}/etc/systemd/network/
```
kengsooncheah, line 18: Is this required? You will be creating symlink to it, right?

kengsooncheah: OK, I think you are creating that directory

kengsooncheah (approved):

PR #85, opened 2022-12-28
Update RCU Hostname Script

HatsyRei:
PR to update rcuhostname script according to strategy listed in [_Rcu_Hostname.md](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/docs/specs/chassis_controller/SmartRacks/RCU_Software/_Rcu_Hostname.md).

Testing done on Gamora, script verified to be able to read from both json files as well as EEPROM:

image

### DIFF
--- a/recipes-core/rcu-hostname/files/rcuhostname
+++ b/recipes-core/rcu-hostname/files/rcuhostname
@@ -4,7 +4,7 @@
 
 hostname=null
 
-# Try to grab custom hostname from mfgconfig.json or .rackconfig.json
+# Try to grab custom hostname from .rackconfig.json
 if [ -f /data/rcu-service/config/.rackconfig.json ]; then
     hostname=$(jq -r .SETTINGS[0].Hostname /data/rcu-service/config/.rackconfig.json)
 fi
--- a/recipes-core/rcu-hostname/files/rcuhostname
+++ b/recipes-core/rcu-hostname/files/rcuhostname
@@ -18,7 +18,7 @@ if [[ $hostname = null ]]; then
 
     if [ -f $EEPROM_SYSFS_PATH  ] && [ $(hexdump -v -e '1/1 "%d"' $EEPROM_SYSFS_PATH -s 0 -n 1) = $EEPROM_REV_REQ ] ; then
 
-        # EEPROM checksum stored at offset 1 byte, total 4 bytes ( -s 1 -n 4 )
+        # EEPROM checksum stored at offset 1, total 4 bytes ( -s 1 -n 4 )
         EEPROM_CHECKSUM=0x$(hexdump -v -e '1/1 "%02x"' $EEPROM_SYSFS_PATH -s 1 -n 4)
 
         # Generate CRC32 from EEPROM content, starting at offset 5, total 981 bytes ( -s 5 -n 981 )
@@ -28,6 +28,8 @@ if [[ $hostname = null ]]; then
 
         # If RCU Carrier EEPROM checksum is valid, read RCU Carrier Module S/N and form hostname
         if [ $EEPROM_CHECKSUM == $EEPROM_CRC32_GEN ]; then
+
+            # RCU Carrier Module S/N stored at offset 412, total 4 bytes ( -s 412 -n 4)
             hostname="ni-rmx-1010x-"$(hexdump -v -e '1/1 "%02x"' $EEPROM_SYSFS_PATH -s 412 -n 4)
         else
             hostname="ni-rmx-1010x"
--- a/recipes-core/rcu-hostname/files/rcuhostname
+++ b/recipes-core/rcu-hostname/files/rcuhostname
@@ -4,11 +4,9 @@
 
 hostname=null
 
-# Try to grab custom hostname from mfgconfig.json or .rackconfig.json
+# Try to grab custom hostname from .rackconfig.json
 if [ -f /data/rcu-service/config/.rackconfig.json ]; then
     hostname=$(jq -r .SETTINGS[0].Hostname /data/rcu-service/config/.rackconfig.json)
-elif [ if /data/rcu-service/config/mfgconfig.json ]; then
-    hostname=$(jq -r .SETTINGS[0].Hostname /data/rcu-service/config/mfgconfig.json)
 fi
 
 # If custom hostname does not exist in JSON config file, try to grab RCU Carrier Module S/N from EEPROM
--- a/recipes-core/rcu-hostname/files/rcuhostname
+++ b/recipes-core/rcu-hostname/files/rcuhostname
@@ -4,7 +4,7 @@
 
 hostname=null
 
-# Try to grab custom hostname from .rackconfig.json
+# Try to grab custom hostname from mfgconfig.json or .rackconfig.json
 if [ -f /data/rcu-service/config/.rackconfig.json ]; then
     hostname=$(jq -r .SETTINGS[0].Hostname /data/rcu-service/config/.rackconfig.json)
 fi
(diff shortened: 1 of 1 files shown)


recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,34 @@
...
 fi
 
-#if serial number is still empty
-if [ -z "$serial" -a "$serial" != " "  ]; then
-    serial="EMPTYSERIAL"
-fi
+# If custom hostname does not exist in JSON config file, try to grab RCU Carrier Module S/N from EEPROM
+if [[ $hostname = null ]]; then
+
+    # If RCU Carrier EEPROM is accessible, verify RCU Carrier EEPROM checksum
```
kengsooncheah: Suggest we also verify the EEPROM revision matches the expected value.

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,34 @@
...
+    if [ -f /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom ]; then
 
-hostname="NI-SmartRacks-"${serial}
+        EEPROM_CHECKSUM=0x$(hexdump -v -e '1/1 "%02x"' /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom -s 1 -n 4)
+        EEPROM_CRC32_GEN=$(python3 -c "import binascii; print(hex(binascii.crc32(b'$(hexdump -v -e '1/1 "\x%02x"' /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom -s 5 -n 981)')))")
```
kengsooncheah: The magic numbers (the offsets and size) may not be obvious to others. Might be good to elaborate it in the comment.

I'm also a little struggle if we should calculate the offset and size based on the "data sets" field because this is the way it's defined. I do see a point why it's not needed here because if the "data sets" isn't equal 5, there is nothing the script can do except for falling to the default hostname, which has the exact effect as failing checksum. If this is your purpose, I think is also good to add a comment about it.

HatsyRei: Yeah.. I think there could be a small benefit in checking the datasets length, since reading 981 bytes from the EEPROM then calculating the checksum in order to fail is more costly. Could save the process like half a second, but again, it's quite the edge case which checksum is supposed to cover in the first place. 

Will make the script clearer in terms of the numbers used.

kengsooncheah: Actually the other problem with calculating checksum according to "data sets" is there's no way for us to verify if "data sets" is valid until it fails the checksum. Imagine if "data sets" equal 0xFF, should we proceed with the CRC calculation (256*196+1 = 50177 bytes)?

The 981 magic number will some what help us to avoid this case. So I'm kind of agree with your way of doing thing (although it's not 100% technically correct) because if the checksum couldn't pass this way, we should just bail out. Just need to provide enough context in the comment for maintenance purpose.

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,34 @@
...
-fi
+hostname=null
 
-#if RCU serial number exists, we prefer the RCU serial number
-if [ -f ${rcuserialpath} ]; then
-    serial=$(tr -d '\0' <${rcuserialpath})
+# Try to grab custom hostname from mfgconfig.json or .rackconfig.json
+if [ -f /data/rcu-service/config/.rackconfig.json ]; then
+    hostname=$(jq -r .SETTINGS[0].Hostname /data/rcu-service/config/.rackconfig.json)
+elif [ if /data/rcu-service/config/mfgconfig.json ]; then
```
karhin1990: Are we going to store hostname into `mfgconfig.json`? I thought this json should only contain immutable info?

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,34 @@
...
 
-hostname="NI-SmartRacks-"${serial}
+        EEPROM_CHECKSUM=0x$(hexdump -v -e '1/1 "%02x"' /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom -s 1 -n 4)
+        EEPROM_CRC32_GEN=$(python3 -c "import binascii; print(hex(binascii.crc32(b'$(hexdump -v -e '1/1 "\x%02x"' /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom -s 5 -n 981)')))")
+
+        # If RCU Carrier EEPROM checksum is valid, read RCU Carrier Module S/N and form hostname
+        if [ $EEPROM_CHECKSUM == $EEPROM_CRC32_GEN ]; then
+            hostname="ni-rmx-1010x-"$(hexdump -v -e '1/1 "%02x"' /sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom -s 412 -n 4)
+        else
+            hostname="ni-rmx-1010x"
```
karhin1990, line 35: Just a thought, if RCU serial number is not available, would it better to append MAC address to the hostname instead of making it a generic `ni-rmx-1010x`? I noticed Linux RT is using MAC address as well when it fails to get SN.

I don't have strong preference though, just want to bring up something to consider. I'll let you and Keng Soon to make the decision 😃

kengsooncheah: I think that's a nice idea.

Anyhow, as far as this PR is concerned, we shall comply to the [spec](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/docs/specs/chassis_controller/SmartRacks/RCU_Software/_Rcu_Hostname.md&_a=preview) that we have already approved.

recipes-core/rcu-hostname/files/rcuhostname
HatsyRei: Ah, you are right. Doesn't seem like we have any use case for custom hostname preloaded by mfg either. I'll remove this part.

recipes-core/rcu-hostname/files/rcuhostname
HatsyRei: Added a check to ensure that revision is as expected. Currently this value is hardcoded to 1 as a variable in the script. Seems like we'll have to update this script as we increment the EEPROM revision later on?

recipes-core/rcu-hostname/files/rcuhostname
HatsyRei: Ok, added comments in one of the latest commits to clarify the numbers.

recipes-core/rcu-hostname/files/rcuhostname
kengsooncheah: Yes, we shall. We shall examine whenever the EEPROM revision changes because different revision could mean different layout and it could be incompatible with this script.

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,40 @@
...
-hostname="NI-SmartRacks-"${serial}
+        # EEPROM checksum stored at offset 1 byte, total 4 bytes ( -s 1 -n 4 )
+        EEPROM_CHECKSUM=0x$(hexdump -v -e '1/1 "%02x"' $EEPROM_SYSFS_PATH -s 1 -n 4)
+
+        # Generate CRC32 from EEPROM content, starting at offset 5, total 981 bytes ( -s 5 -n 981 )
+        # 1 byte at offset 5 stores the number of datasets present in the EEPROM. The expected number of datasets in RCU Carrier EEPROM is 5.
+        # Each dataset is 196 bytes in length. Hence, 1 + (196 * 5) = 981 bytes
+        EEPROM_CRC32_GEN=$(python3 -c "import binascii; print(hex(binascii.crc32(b'$(hexdump -v -e '1/1 "\x%02x"' $EEPROM_SYSFS_PATH -s 5 -n 981)')))")
+
+        # If RCU Carrier EEPROM checksum is valid, read RCU Carrier Module S/N and form hostname
```
kengsooncheah, line 29: Suggest you to add comment for the magic number 412 too

recipes-core/rcu-hostname/files/rcuhostname
```diff
@@ -2,28 +2,40 @@
...
+if [[ $hostname = null ]]; then
+
+    # If RCU Carrier EEPROM is accessible and EEPROM revision is expected, verify RCU Carrier EEPROM checksum
+    EEPROM_SYSFS_PATH=/sys/devices/platform/bus@56240000/56246000.i2c/i2c-*/*-0050/eeprom
+    EEPROM_REV_REQ=1
+
+    if [ -f $EEPROM_SYSFS_PATH  ] && [ $(hexdump -v -e '1/1 "%d"' $EEPROM_SYSFS_PATH -s 0 -n 1) = $EEPROM_REV_REQ ] ; then
 
-hostname="NI-SmartRacks-"${serial}
+        # EEPROM checksum stored at offset 1 byte, total 4 bytes ( -s 1 -n 4 )
```
kengsooncheah: Suggest you to make the terms consistent, offset is normally counted at byte-boundary, hence "offset 1" should be sufficient.

kengsooncheah (approved):

karhin1990 (approved):

PR #88, opened 2023-01-17
Persist RCU user credentials across Mender updates

HatsyRei:
PR to address [US 2172148](https://dev.azure.com/ni/DevCentral/_workitems/edit/2172148).

Mender state script content largely leveraged from examples provided here:
https://github.com/mendersoftware/mender/tree/master/examples/state-scripts

Technique:
Since symlinking /etc/passwd, /etc/shadow, etc in /data is [discouraged](https://serverfault.com/questions/491033/), the only solution left is to manually copy them into the new rootfs partition after the new RCU image has been written to the target partition by Mender. To achieve this, [state scripts](https://docs.mender.io/artifact-creation/state-scripts) are utilized to mount the target partition and copy over the relevant credential files.

Reference:
https://hub.mender.io/t/maintaining-linux-unique-password/3244
https://hub.mender.io/t/rootfs-update-module-keep-existing-file-from-active-partition-in-new-partition/2465
https://learning.lpi.org/en/learning-materials/010-160/5/5.2/5.2_01/

Testing:
Solution verified on RCU + fanout board. Root account password modified using `passwd` command. Modified password still applies after Mender update to new RCU image containing the added state script.

### DIFF
--- a/recipes-mender/rcu-state-scripts/files/retain-credentials
+++ b/recipes-mender/rcu-state-scripts/files/retain-credentials
@@ -40,6 +40,7 @@ if [ -d /mnt/etc ]; then
     echo "Copied user and group credentials to new root partition" >&2
 else
     echo "Failed to find /etc on new root partition" >&2
+    umount $newroot
     exit 1
 fi
 
--- a/recipes-mender/rcu-state-scripts/files/retain-credentials
+++ b/recipes-mender/rcu-state-scripts/files/retain-credentials
@@ -5,13 +5,15 @@
 
 echo "$(mender show-artifact): Running $(basename "$0")" >&2
 
-# Testing 
+# Check if fw_printenv command is available
 if [ ! -x /usr/bin/fw_printenv ]; then
     exit 1
 fi
 
+# Check current rootfs partition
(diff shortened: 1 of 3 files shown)


recipes-mender/rcu-state-scripts/files/retain-credentials
```diff
@@ -0,0 +1,46 @@
...
+
+if [ -d /mnt/etc ]; then
+    cp /etc/passwd /mnt/etc/passwd
+    cp /etc/shadow /mnt/etc/shadow
+    cp /etc/group /mnt/etc/group
+    cp /etc/gshadow /mnt/etc/gshadow
+    echo "Copied user and group credentials to new root partition" >&2
+else
+    echo "Failed to find /etc on new root partition" >&2
+    exit 1
```
kengsooncheah, line 44: Should we do an exit here? `$newroot` will not get unmounted.

HatsyRei: Ah, good catch. The `exit 1` will cause Mender to not proceed with the update, so I suppose we'd still want to exit 1, but also make sure we unmount before this statement.

kengsooncheah (approved):

PR #91, opened 2023-03-02
Install ssl cert

shlooiNI:
### Justification
Update recipe to install SSL/TLS certificate onto RCU.

### Changes
Update recipe to install SSL/TLS public certificate onto RCU.

### Note
This PR is dependent on https://github.com/ni/smartracks-script/pull/32

### DIFF
--- a/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
+++ b/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
@@ -17,9 +17,4 @@ do_install() {
          install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
          install -d ${D}/${sysconfdir}/ssl/certs
          install -m 0644 ${S}/certs/ni_ate_core_cert.pem ${D}/${sysconfdir}/ssl/certs
-<<<<<<< HEAD
-=======
-         install -d ${D}/${sysconfdir}/ssl/private
-         install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
->>>>>>> Change permissions for private key
 }
--- a/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
+++ b/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
@@ -17,4 +17,9 @@ do_install() {
          install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
          install -d ${D}/${sysconfdir}/ssl/certs
          install -m 0644 ${S}/certs/ni_ate_core_cert.pem ${D}/${sysconfdir}/ssl/certs
+<<<<<<< HEAD
+=======
+         install -d ${D}/${sysconfdir}/ssl/private
+         install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
+>>>>>>> Change permissions for private key
 }
--- a/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
+++ b/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
@@ -15,4 +15,6 @@ do_install() {
          install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
          install -d ${D}/${sysconfdir}/ssl/private
          install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
+         install -d ${D}/${sysconfdir}/ssl/certs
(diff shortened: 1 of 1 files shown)


recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
```diff
@@ -13,5 +13,9 @@ do_install() {
          install -m 0755 rcu-service ${D}${bindir}
          install -d ${D}/${systemd_unitdir}/system
          install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
+         install -d ${D}/${sysconfdir}/ssl/certs
+         install -m 0644 ${S}/certs/ni_ate_core_cert.pem ${D}/${sysconfdir}/ssl/certs
+         install -d ${D}/${sysconfdir}/ssl/private
+         install -m 0644 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
```
kengsooncheah: we do not want to grant others even the read permission. I think the mode shall be 0640. @HatsyRei can help to confirm?

HatsyRei: Ah, yes. 0640 should work.
https://chmodcommand.com/chmod-640/

kengsooncheah: Maybe one thing to think about the pem and private key. Do we want to store it with our repo? Or should we only inject it in the mfg process.

shlooiNI: Updated to 640

kengsooncheah: I checked the private key and ownership of my Ubuntu Linux machine, it seems like there is a special owner for the directory

```
kencheah@rog-1:~/mnt/smartrack/repo/archive$ sudo ls -al /etc/ssl/private/
total 12
drwx--x--- 2 root ssl-cert 4096 Nov  12  2021 .
drwxr-xr-x 4 root root     4096 Feb   8 07:00 ..
-rw-r----- 1 root ssl-cert 1708 Nov  12  2021 ssl-cert-snakeoil.key
```

shlooiNI: > Maybe one thing to think about the pem and private key. Do we want to store it with our repo? Or should we only inject it in the mfg process.

pem is ok to store, as it is supposed to be public anyway. @HatsyRei did suggest we can look into using the private tokens function in github to maybe store the private key. For now I have it in our repo, as our repo is private it should be ok for now. But will look into maybe using the private token stuff

kengsooncheah: The tricky part with git is once you commit, it will stay there forever, even if you remove it later. So we have to think through the plan to manage the private key.

karhin1990 (approved):

recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
shlooiNI: That is true, but we can always regenerate a new key, which would make the old ones invalid if that is a concern.

HatsyRei: > I checked the private key and ownership of my Ubuntu Linux machine, it seems like there is a special owner for the directory
> 
> ```
> kencheah@rog-1:~/mnt/smartrack/repo/archive$ sudo ls -al /etc/ssl/private/
> total 12
> drwx--x--- 2 root ssl-cert 4096 Nov  12  2021 .
> drwxr-xr-x 4 root root     4096 Feb   8 07:00 ..
> -rw-r----- 1 root ssl-cert 1708 Nov  12  2021 ssl-cert-snakeoil.key
> ```

@kengsooncheah Hmm, seems like we don't have this under RCU. Private dir also belongs to root.

kengsooncheah: ya... not sure if there's any best practices to handle the security on a yocto-based device. I'm trying to search if can find anything...

nwilsonni: > Maybe one thing to think about the pem and private key. Do we want to store it with our repo? Or should we only inject it in the mfg process.

probably should not be part of open source section of code

shlooiNI:
@HatsyRei @kengsooncheah @karhin1990 I've updated this PR, since we already merged in the changes to install private key.

HatsyRei (approved):

recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
shlooiNI: The private key and certs are now stored as AzDo secure files, and are not part of our github repository.

kengsooncheah (approved):

PR #92, opened 2023-03-03
Add script to reset RCU IP configuration

HatsyRei:
Add a simple script to reset RCU IP configuration to default configuration. Useful to SW test and can potentially be leveraged by rcu-service when implementing IP reset button.

### DIFF
--- a/recipes-core/systemd/files/reset-ip-config
+++ b/recipes-core/systemd/files/reset-ip-config
@@ -18,7 +18,6 @@ DHCP=ipv4
 " > /data/systemd/network/eth0.network
 
 echo "
-
 [Match]
 Name=eth1
 
--- a/recipes-core/systemd/files/reset-ip-config
+++ b/recipes-core/systemd/files/reset-ip-config
@@ -23,8 +23,6 @@ Name=eth1
 
 [Network]
 Address=10.1.1.20/24
-Gateway=10.1.1.1
-DNS=10.1.1.1
 " > /data/systemd/network/eth1.network
 
 sync
--- a/recipes-core/systemd/files/reset-ip-config
+++ b/recipes-core/systemd/files/reset-ip-config
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# Script which resets RCU IP address configuration for both eth0 and eth1
+# This is done by
+# 1) Removing eth0 and eth1 network configuration files from /data/systemd/network
+# 2) Recreating those files with default eth0 and eth1 configuration
+# 3) Restarting systemd-networkd for changes to take effect
+
+rm /data/systemd/network/eth0.network
+rm /data/systemd/network/eth1.network
+
+echo "
+[Match]
+Name=eth0
+
+[Network]
+DHCP=ipv4
+" > /data/systemd/network/eth0.network
+
+echo "
+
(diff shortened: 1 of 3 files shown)


recipes-core/systemd/files/reset-ip-config
```diff
@@ -0,0 +1,32 @@
...
+[Network]
+DHCP=ipv4
+" > /data/systemd/network/eth0.network
+
+echo "
+[Match]
+Name=eth1
+
+[Network]
+Address=10.1.1.20/24
```
shlooiNI, line 25: Doesn't setting the Gateway and DNS cause issues with the connection?

recipes-core/systemd/files/reset-ip-config
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
shlooiNI, line 1: Can we walso reset the rackconfig or mfgconfig json file? I'm concerned that in this case the config json may still have the old settings in them.

recipes-core/systemd/files/reset-ip-config
HatsyRei: Oh yeah, I think we'll need to change this and also the default eth1.network file... we haven't done that yet. I think it should be fine to do that here.

recipes-core/systemd/files/reset-ip-config
HatsyRei: Ah, I see what you mean here. This could be difficult to do in the script itself since the norm we have established is that the only one which can write the config files is rcu-service- Preferably we keep it that way, so I think what needs to happen when IP reset button is pressed is, we call this script and then we need to invoke the `getIpConfigFromNetworkFile` and `setIpConfig` calls to update the config json. Will this be hard to do in rcu-service?

As for test engineer usage during FVT... they might call this script directly, so I'll need to make sure that they clear out the .rackconfig json file after the test. Should be part of the test steps since if not mistaken they will populate new mfgconfig,json during frame/system level test.

shlooiNI: I think it should be quite trivial to call those functions from rcu-service. But we'll need to decide where and how to handle it. 
During the test, how will this script be invoked? Will it be through the IP-Reset button? If so, we can invoke the `setIpConfig` functions after the button is pressed and the files are updated like you suggested.

HatsyRei: During RCU Carrier FVT, rcu-service will be in halt state, so almost all of their test steps will be on command line. They will call the script directly from command line and won't go through rcu-service. They will manually clean up the /data/rcu-service/config directory after the test.

recipes-core/systemd/files/reset-ip-config
HatsyRei: Updated in latest commits.

recipes-core/systemd/files/reset-ip-config
shlooiNI: Ok. If the config files are cleaned up it should be ok. Because upon reboot, rcu-service should be able to detect that the config files do not have the eth settings and will populate it.

shlooiNI (approved):

PR #93, opened 2023-03-07
Disable SSH server by default and persist setting across updates

HatsyRei:
PR to disable dropbear SSH server from running by default on RCU. Also added mender state scripts which will persist SSH server availability across mender updates. This allows the following behavior to take place:

1. If SSH server is currently enabled on RCU, mender update will not disable SSH server.
2. If SSH server is already disabled on RCU, SSH server remains disabled after mender update.
3. Flashing RCU via TEZI always disables RCU SSH server by default. 

Testing done on local RCU setup to verify the behavior stated above.


shlooiNI (approved):
Looks good to me.

kengsooncheah (approved):

PR #96, opened 2023-03-15
Test using private key stored as secret

shlooiNI:
## Justification
This is to serve as a test to prove that we can reference github secrets into files, that will then be packaged into mender image. If this works, it may be how we inject the gRPC TLS private key into RCU image.

## Changes
Echo secret private key into key file before packaging it into mender image.
Add steps to install private key onto RCU.

### DIFF
--- a/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
+++ b/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
@@ -9,7 +9,7 @@ SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
 do_configure_append() {
-    echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
+         echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
 }
 
 do_install() {
--- a/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
+++ b/recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
@@ -8,7 +8,9 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
-echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
+do_configure_append() {
+    echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
+}
 
 do_install() {
          install -d ${D}${bindir}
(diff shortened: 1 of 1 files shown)


recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
```diff
@@ -8,10 +8,14 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
+echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
```
HatsyRei: I tried something like this, will probably result in a ParseError. Suggesting to try put it in a do_configure_append() block to fix the issue:

```
do_configure_append() {
    echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
}
``` 

I was able to get a test file installed this way, so it comes down to whether the NI_ATE_CORE_PRIVATE_KEY variable works correctly. Good luck! :D

shlooiNI: Thanks!

recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
```diff
@@ -8,10 +8,16 @@ inherit systemd
 SYSTEMD_AUTO_ENABLE = "enable"
 SYSTEMD_SERVICE_${PN} = "rcu-service.service"
 
+do_configure_append() {
+    echo $NI_ATE_CORE_PRIVATE_KEY > ${S}/certs/ni_ate_core_private.key
```
kengsooncheah: Let's align the indentation

recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
```diff
@@ -8,10 +8,16 @@ inherit systemd
...
+}
+
 do_install() {
          install -d ${D}${bindir}
          install -m 0755 rcu-service ${D}${bindir}
          install -d ${D}/${systemd_unitdir}/system
          install -m 0644 ${S}/rcu-service.service ${D}/${systemd_unitdir}/system
+         install -d ${D}/${sysconfdir}/ssl/private
+         install -m 0640 ${S}/certs/ni_ate_core_private.key ${D}/${sysconfdir}/ssl/private
```
kengsooncheah, line 23: When I think twice on our plan to submit the PR which include a private key and remove it, probably let's just rework a fresh one with no private key at all to avoid unnecessary problems. We can maybe prepare a blank file if we want.

kengsooncheah (approved):

recipes-ni/proprietary/rcu-service/rcu-service_1.0.bb
shlooiNI: Are you referring to https://github.com/ni/rcu-service/pull/245 ?

kengsooncheah: Yes

shlooiNI: Created a new PR here https://github.com/ni/rcu-service/pull/256

PR #101, opened 2023-03-23
Update reset IP script to update json file

WayneDroid:
__LATEST_UPDATE:__
We decide to move the responsibility of network setting update on rackjson file to rcu-service, so this PR is no longer needed.

__Justification:__
We need to reset network setting entry in /data/rcu-service/config/.rackconfig.json when we reset .network files in /data/systemd/network/.

__Changes:__
- using jq to modify .rackconfig.json file.

__Testing:__
Tested on Widow, the .rackconfig.json is updated with default network setting entry.

AzDO work item: https://dev.azure.com/ni/DevCentral/_workitems/edit/2312260

### DIFF
--- a/recipes-core/systemd/files/reset-ip-config
+++ b/recipes-core/systemd/files/reset-ip-config
@@ -14,7 +14,7 @@
 RACK_CONFIG_FILE_PATH="/data/rcu-service/config/.rackconfig.json"
 RACK_CONFIG_BACKUP_FILE_PATH="/data/rcu-service/config/.rackconfig-backup.json"
 # default ethernet settings
-DEFAULT_SETTING='"SETTINGS":[{"ID":"Settings","Eth0":"DHCP","Eth1":"10.1.1.20/24::10.1.1.1::10.1.1.1"}]}'
+DEFAULT_SETTING='"SETTINGS":[{"ID":"Settings","Eth0":"DHCP","Eth1":"10.1.1.20/24"}]}'
 
 validate_json_file()
 {
--- a/recipes-core/systemd/files/reset-ip-config
+++ b/recipes-core/systemd/files/reset-ip-config
@@ -5,6 +5,33 @@
 # 1) Removing eth0 and eth1 network configuration files from /data/systemd/network
 # 2) Recreating those files with default eth0 and eth1 configuration
 # 3) Restarting systemd-networkd for changes to take effect
+# 4) Making copy of .rackconfig.json to .rackconfig-backup.json
+# 5) Remove SETTINGS entry if it exists
+# 6) Add default SETTINGS entry and save to .rackconfig.json
+# 7) Revert .rackconfig.json if error.
+# 8) Remove .rackconfig-backup.json
+
+RACK_CONFIG_FILE_PATH="/data/rcu-service/config/.rackconfig.json"
+RACK_CONFIG_BACKUP_FILE_PATH="/data/rcu-service/config/.rackconfig-backup.json"
+# default ethernet settings
+DEFAULT_SETTING='"SETTINGS":[{"ID":"Settings","Eth0":"DHCP","Eth1":"10.1.1.20/24::10.1.1.1::10.1.1.1"}]}'
+
+validate_json_file()
+{
+    strVar1=$(echo $rackconfigVar | jq -c '.')
+    strVar2=$(cat $RACK_CONFIG_FILE_PATH | jq -c '.')
+    
+    if [ "$strVar1" != "$strVar2" ]; then
+        # rename backup file to original file
+        echo "update config file fails"
+        rm $RACK_CONFIG_FILE_PATH
+        mv $RACK_CONFIG_BACKUP_FILE_PATH $RACK_CONFIG_FILE_PATH
+        exit 1
+    else
+        echo "removing .rackconfig-backup.json"
+        rm $RACK_CONFIG_BACKUP_FILE_PATH
+    fi
+}
 
 rm /data/systemd/network/eth0.network
 rm /data/systemd/network/eth1.network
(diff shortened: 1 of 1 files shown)


recipes-core/systemd/files/reset-ip-config
```diff
@@ -5,26 +5,75 @@
...
+# 4) Making copy of .rackconfig.json to .rackconfig-backup.json
+# 5) Remove SETTINGS entry if it exists
+# 6) Add default SETTINGS entry and save to .rackconfig.json
+# 7) Revert .rackconfig.json if error.
+# 8) Remove .rackconfig-backup.json
+
+RACK_CONFIG_FILE_PATH="/data/rcu-service/config/.rackconfig.json"
+RACK_CONFIG_BACKUP_FILE_PATH="/data/rcu-service/config/.rackconfig-backup.json"
+# default ethernet settings
+DEFAULT_SETTING='"SETTINGS":[{"ID":"Settings","Eth0":"DHCP","Eth1":"10.1.1.20/24::10.1.1.1::10.1.1.1"}]}'
```
shlooiNI: I think the default Eth1 settings might need to be different as I believe setting the Gateway may cause issues. @HatsyRei can you advise?

HatsyRei: No DNS and gateway in eth1 static IP specification. Just

`"SETTINGS":[{"ID":"Settings","Eth0":"DHCP","Eth1":"10.1.1.20/24"}]}`

if I am not mistaken.

WayneDroid: @HatsyRei I will update it.

HatsyRei:
@WayneDroid Will there be a problem if rcu-service and the script writes to .rackconfig.json simultaneously? E.g., user sends gRPC API to set fan speed, and reset IP script is invoked at the same time. Looks like two separate processes will modify the .json file.

WayneDroid:
> @WayneDroid Will there be a problem if rcu-service and the script writes to .rackconfig.json simultaneously? E.g., user sends gRPC API to set fan speed, and reset IP script is invoked at the same time. Looks like two separate processes will modify the .json file.

@HatsyRei 
I never thought of the situation on two processes happen same time. Anyhow, the network setting will be updated in .rackconfig.json file when user triggers IP reset on the Reset button. I am not sure how likely would user resets network while another one setting the fan speed.
Do you have any idea how to handle this situation? Or we let rcu-service to modify the network setting in .rackconfig.json file?

HatsyRei:
@WayneDroid 

Looks like this really isn't easy to resolve.. I gave it some thought and I feel that having rcu-service call the IP reset script may not be the best way to do things. Originally, the IP reset script was created because test team wanted an easy way to reset IP during/after RCU FVT. I thought that the script can be reused by rcu-service to reset IP configuration, but when factoring in how JSON config file also needs to be reset, I think the script will not be able to do a proper job. 

I would suggest omitting the script from our IP reset implementation, and have rcu-service completely handle resetting IP configuration in both .network and .json files. On rcu-service side, we can leverage [`tIpConfigManager::setIpConfig`](https://github.com/ni/rcu-service/blob/main/source/utils/tIpConfigManager.cpp#L39) to directly update both .network and .json using a hardcoded static IP configuration. Do you think this is a good idea?

kengsooncheah:
> @WayneDroid
> 
> Looks like this really isn't easy to resolve.. I gave it some thought and I feel that having rcu-service call the IP reset script may not be the best way to do things. Originally, the IP reset script was created because test team wanted an easy way to reset IP during/after RCU FVT. I thought that the script can be reused by rcu-service to reset IP configuration, but when factoring in how JSON config file also needs to be reset, I think the script will not be able to do a proper job.
> 
> I would suggest omitting the script from our IP reset implementation, and have rcu-service completely handle resetting IP configuration in both .network and .json files. On rcu-service side, we can leverage [`tIpConfigManager::setIpConfig`](https://github.com/ni/rcu-service/blob/main/source/utils/tIpConfigManager.cpp#L39) to directly update both .network and .json using a hardcoded static IP configuration. Do you think this is a good idea?

I actually share the same concern. I start to become nervous if someone other than rcu-service can write to .rackconfig.json.

WayneDroid:
> @WayneDroid
> 
> Looks like this really isn't easy to resolve.. I gave it some thought and I feel that having rcu-service call the IP reset script may not be the best way to do things. Originally, the IP reset script was created because test team wanted an easy way to reset IP during/after RCU FVT. I thought that the script can be reused by rcu-service to reset IP configuration, but when factoring in how JSON config file also needs to be reset, I think the script will not be able to do a proper job.
> 
> I would suggest omitting the script from our IP reset implementation, and have rcu-service completely handle resetting IP configuration in both .network and .json files. On rcu-service side, we can leverage [`tIpConfigManager::setIpConfig`](https://github.com/ni/rcu-service/blob/main/source/utils/tIpConfigManager.cpp#L39) to directly update both .network and .json using a hardcoded static IP configuration. Do you think this is a good idea?

@HatsyRei @kengsooncheah Yeah, I agree with you, we should only update both .network and json files from rcu-service, `tIpConfigManager.cpp` is a best place to reset the updates.
Anyhow, we still can keep this script to accommodate test team's request. As long as we don't call it from rcu-service. What do you think?

nwilsonni:
If we have concerns about Resetting the IP address when something else may be going on with the system, we can create step in the Integration Guide to say that when resetting IP or PW, the system should be powered off and not have any active remote connections (we can work on the wording)   I assume they will need to reboot the RCU, power cycle the rack, after the change anyway.

WayneDroid:
We decide to move the responsibility of network setting update on rackjson file to rcu-service, so this PR is no longer needed.

PR #102, opened 2023-04-03
Add crypt dependencies to rcu-service recipe

HatsyRei:
OE builds complaining about missing crypt.h after https://github.com/ni/rcu-service/pull/266/. This adds the proper dependencies to allow crypt library to be included in rcu-service OE build.

See https://www.yoctoproject.org/pipermail/yocto/2019-May/045481.html


shlooiNI (approved):
Thanks for quick fix!

PR #103, opened 2023-04-03
Retain private keys, remove private key installation steps

shlooiNI:
## Justification
Tied to this [User Story](https://dev.azure.com/ni/DevCentral/_workitems/edit/2352123). This is part of a more secure proposal on how to inject the SSL/TLS private key into RCU. 
Current method is to package private key in RCU image and install it to RCU during firmware updates. This is not secure as the RCU image can be unzipped easily to obtain the private key.
New proposal is the have private key injected during RCU FVT.

## Changes
1. Add script to retain private keys during mender updates.
2. Remove installation step for private key.

### DIFF
--- a/recipes-mender/rcu-state-scripts/files/retain-ssl-private-keys
+++ b/recipes-mender/rcu-state-scripts/files/retain-ssl-private-keys
@@ -33,7 +33,7 @@ fi
 sleep 2
 
 if [ -d /mnt/etc ]; then
-    cp /etc/ssl/private /mnt/etc/ssl/private
+    cp -r /etc/ssl/private /mnt/etc/ssl/private
     echo "Copied private keys to new root partition" >&2
 else
     echo "Failed to find /etc on new root partition" >&2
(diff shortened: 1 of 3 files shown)


recipes-mender/rcu-state-scripts/files/retain-ssl-private-keys
```diff
@@ -0,0 +1,44 @@
...
+
+if [ $? -ne 0 ]; then
+    echo "Failed to mount $newroot" >&2
+    exit 1
+fi
+
+sleep 2
+
+if [ -d /mnt/etc ]; then
+    cp /etc/ssl/private /mnt/etc/ssl/private
```
HatsyRei: Are we copying dir contents? Do we need `cp -r`?

https://superuser.com/questions/1418605/

shlooiNI: Oh good point. Yes I intend to copy the dir contents in case users have more than one key stored.

HatsyRei (approved):

kengsooncheah:
I think we need to repeat the same step for public cert, right? They always come in pair.

kengsooncheah (approved):

PR #104, opened 2023-04-05
Hotfix for retain ssl private key script

shlooiNI:
## Justification
Hotfix for retain ssl private key script as it was incorrectly adding the entire private folder into /etc/ssl/private

## Changes
Use copy of folder contents instead of cp -r


kengsooncheah (approved):
This is interesting. Any idea why `cp -r` didn't work?

shlooiNI:
> This is interesting. Any idea why `cp -r` didn't work?

@kengsooncheah the `cp -r` was copying the entire directory eg `private` as a folder, and copying it into `etc/ssl/private` causing the end path to be `etc/ssl/private/private/`

PR #107, opened 2023-04-17
Add Mender state scripts to retain dbus machine-id

HatsyRei:
Previously, RCU Mender updates were found to cause RCU IP to change. This causes issues for test automation, as the DNS server records need to be updated and the DNS client needs to refresh its cache to correctly connect to the RCU via hostname.

After investigation, it is found that the dbus machine-id is used for the creation of DHCP host identifier [1]. Machines having the same machine-id can even obtain the same DHCP IP address from a DHCP server, something which happens frequently on cloned VMs [2].

This PR introduces Mender state scripts to retain the /etc/machine-id file across Mender updates. Changes tested on local RCU setup and IP address was observed to not change from update to update after the machine-id file was retained successfully.

[1] https://wiki.debian.org/MachineId
[2] https://superuser.com/questions/1380443/

### DIFF
--- a/recipes-mender/rcu-state-scripts/files/retain-dbus-machine-id
+++ b/recipes-mender/rcu-state-scripts/files/retain-dbus-machine-id
@@ -1,0 +1,44 @@
+#!/bin/sh
+#
+# Retain dbus machine-id file from the current root
+#
+
+echo "$(mender show-artifact): Running $(basename "$0")" >&2
+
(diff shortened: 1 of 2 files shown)


shlooiNI (approved):

kengsooncheah (approved):

nwilsonni:
ah - very NICE to see this - it was always a pain when IP would change after image deploy!!!!

PR #109, opened 2023-05-09
Fix build dependency issues for building RCU pytest

HatsyRei:
Follow up to https://github.com/ni/meta-smartracks/pull/108. Saw pipeline build failures complaining about bdist_wheel not being a valid command. Reproduced failures and tested fix on local build machine

### DIFF
--- a/recipes-ni/proprietary/rcu-service/rcu-service-src.inc
+++ b/recipes-ni/proprietary/rcu-service/rcu-service-src.inc
@@ -13,7 +13,7 @@ SRCREV = "${AUTOREV}"
 
 S = "${WORKDIR}/git"
 
-DEPENDS += " python3-grpcio-tools-native virtual/crypt "
+DEPENDS += " python3-grpcio-tools-native virtual/crypt python3-wheel-native "
 RDEPENDS_${PN} += " python3 python3-grpcio "
 inherit python3native
 inherit python3-dir
--- a/recipes-ni/proprietary/rcu-service/rcu-service-src.inc
+++ b/recipes-ni/proprietary/rcu-service/rcu-service-src.inc
@@ -13,7 +13,7 @@ SRCREV = "${AUTOREV}"
 
 S = "${WORKDIR}/git"
 
-DEPENDS += " python3-grpcio-tools-native virtual/crypt python3-wheel-native "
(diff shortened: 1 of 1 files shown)


recipes-ni/proprietary/rcu-service/rcu-service-src.inc
```diff
@@ -13,7 +13,7 @@ SRCREV = "${AUTOREV}"
 
 S = "${WORKDIR}/git"
 
-DEPENDS += " python3-grpcio-tools-native virtual/crypt "
+DEPENDS += " python3-grpcio-tools-native virtual/crypt python3-wheel-native "
```
kengsooncheah: Ah, OK. I thought the container has the dependencies needed. Besides wheel, the other dependency we need is setuptools as mentioned in the [readme](https://github.com/ni/rcu-service/tree/main/tests/pytest#how-to-build-the-python-package-wheel-build-id). Do you know if that requirement is already satisfied?

HatsyRei: I tried adding both `python3-setuptools-native` and `python3-wheel-native`, which allowed the build to pass. I then removed `python3-setuptools-native` and realized that the build was still fine, seemingly narrowing down the missing package to `python3-wheel-native`. 

I'll give this a few more tries and see if it is passing due to some cached files.

shlooiNI (approved):

recipes-ni/proprietary/rcu-service/rcu-service-src.inc
HatsyRei: Added `python3-setuptools-native`. Thinking it is better to spell it out explicitly.

kengsooncheah (approved):

recipes-ni/proprietary/rcu-service/rcu-service-src.inc
kengsooncheah: Sounds good. I tried to find out if `setuptools` is implicitly included in `wheel`, but couldn't find much info. The [`wheel` documentation](https://wheel.readthedocs.io/en/stable/) mentioned it is a `setuptools` extension. Not sure what it means exactly but I guess it's extends `setuptools` instead of providing `setuptools`. I guess `setuptools` is probably installed in the base image or being required by other modules. We need `setuptools` to build the wheel because it is used in our code, so I think no harm including it.

PR #110, opened 2023-05-22
Accomodate repo name changes to 'ateccgen2'

HatsyRei:
Just an initial look at how much work needs to be done to transition to using 'ateccgen2' instead of 'smartracks'. Will be doing the same for other repos. Not intended to be merged yet.


recipes-images/images/smartracks-minimal-image.bb
kengsooncheah, line 1: I kind of curious why git did not detect this file as moved. I guess there are 2 kind of changes happened at the same time, 1.) rename of the file, 2.) modification of the content. Did you use `git mv` to rename the file?

stub.yml
```diff
@@ -1,9 +1,9 @@
 # Stub pipeline
-# Updates the TriggeringPipeline variable in SmartRacksPipelineVariables group.
+# Updates the TriggeringPipeline variable in AteCoreConfigGen2PipelineVariables group.
 # After execution of this pipeline is finished, rcu-image-ci-pipeline.yml located in smatracks-script should be executed.
```
kengsooncheah: Same typo that we need to change the name.

recipes-images/images/smartracks-minimal-image.bb
HatsyRei: Yeah, the modification was to change the `require smartracks-base-image.inc` statement to `require ateccgen2-base-image.inc`. I did not use the `git mv` command here.

PR #120, opened 2023-07-24
Add Mender update sanity check service

HatsyRei:
PR to add a post-Mender-update RCU sanity check script. The current logic is primitive- On image installation, /etc/rcu-mender-image-firstboot file is installed. This file serves as a flag for RCU to run the sanity check service, which removes the flag, then checks if rcu-service is running. If rcu-service is running, mender commit, flag a successful update and restart rcu-service so the flag gets read. If rcu-service fails to run, mender rollback (and wait for user to power cycle the rack/RCU).

Flags which will be interpreted by rcu-service for deducing last update status (pending code update):
/data/rcu-service/firmware/update-staged
/data/rcu-service/firmware/update-failed
/data/rcu-service/firmware/update-succeeded

There is a chance that an image is faulty and not even able to boot to userspace- The sanity check script will not even be executed. In this case, the plan is that once the RCU boots to the old image after an automatic reboot or a power cycle, rcu-service will still be able to interpret the existence of 'update-staged' flag and deduce that the sanity check script failed to remove this.

Another potential case to handle is how the script behaves on TEZI installation- As the script is right now, it will run but will not locate the 'update-staged' flag, nor will mender commit or rollback commands work. However, this should not cause any issues. A way to address this could be to only install the /etc/rcu-mender-image-firstboot file via Mender state scripts- That will require some additional work to get right. 

Some additional checks I would like to include in this involves checking if the RCU image version is as expected after an update- This again requires some work on rcu-service side. We can add it in when the rcu-service code is ready.

### DIFF
--- a/recipes-core/rcu-sanity-check/files/post-update-sanity-check
+++ b/recipes-core/rcu-sanity-check/files/post-update-sanity-check
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+#rm /data/rcu-service/firmware/rcu-update-staged
+
+# TODO: Check if network is up
+# TODO: Check if rcu-service is active
+
+# If sanity check completes successfully, mender commit and set flag
+#mender commit
+#touch /data/rcu-service/firmware/rcu-update-success
+
+# If sanity check fails, mender rollback, set flag and reboot
+#mender rollback
+#touch /data/rcu-service/firmware/rcu-update-failed
+#reboot
--- a/recipes-core/rcu-sanity-check/files/post-update-sanity-check
+++ b/recipes-core/rcu-sanity-check/files/post-update-sanity-check
@@ -1,15 +1,16 @@
 #!/bin/sh
 
-#rm /data/rcu-service/firmware/rcu-update-staged
+rm /data/rcu-service/firmware/update-staged
+rm /etc/rcu-mender-image-firstboot
 
-# TODO: Check if network is up
-# TODO: Check if rcu-service is active
+RCU_SERVICE_STATUS=$(systemctl is-active rcu-service)
+ETH0_STATUS=$(cat /sys/class/net/eth0/operstate)
 
-# If sanity check completes successfully, mender commit and set flag
-#mender commit
-#touch /data/rcu-service/firmware/rcu-update-success
-
-# If sanity check fails, mender rollback, set flag and reboot
-#mender rollback
-#touch /data/rcu-service/firmware/rcu-update-failed
-#reboot
+if [ "$RCU_SERVICE_STATUS" = "active" ]; then
+    mender commit
+    touch /data/rcu-service/firmware/update-succeeded
+    systemctl restart rcu-service
+else
+    mender rollback
+    touch /data/rcu-service/firmware/update-failed
+fi;
--- a/recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
+++ b/recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Service to run RCU sanity check
+After=network.target
+ConditionPathExists=/data/rcu-service/firmware/rcu-update-staged
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/post-update-sanity-check
+
+[Install]
+WantedBy=multi-user.target
--- a/recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
+++ b/recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Service to run RCU sanity check
 After=network.target
-ConditionPathExists=/data/rcu-service/firmware/rcu-update-staged
+ConditionPathExists=/etc/rcu-mender-image-firstboot
 
 [Service]
 Type=oneshot
--- a/recipes-core/rcu-sanity-check/rcu-sanity-check_1.0.bb
+++ b/recipes-core/rcu-sanity-check/rcu-sanity-check_1.0.bb
@@ -8,11 +8,13 @@ inherit systemd
 SRC_URI += " \
     file://post-update-sanity-check \
     file://post-update-sanity-check.service \
+    file://rcu-mender-image-firstboot \
 "
 
 FILES:${PN} = " \
     ${bindir} \
     ${systemd_system_unitdir} \
+    ${sysconfdir} \
 "
 
 SYSTEMD_SERVICE:${PN} = " \
@@ -24,4 +26,6 @@ do_install () {
     install -m 0755 ${WORKDIR}/post-update-sanity-check ${D}${bindir}
     install -d ${D}${systemd_system_unitdir}
     install -m 0644 ${WORKDIR}/post-update-sanity-check.service ${D}${systemd_system_unitdir}
+    install -d ${D}${sysconfdir}
+    install -m 0644 ${WORKDIR}/rcu-mender-image-firstboot ${D}${sysconfdir}
 }
--- a/recipes-core/rcu-sanity-check/rcu-sanity-check_1.0.bb
+++ b/recipes-core/rcu-sanity-check/rcu-sanity-check_1.0.bb
@@ -1,0 +1,27 @@
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit systemd
+
+SRC_URI += " \
+    file://post-update-sanity-check \
+    file://post-update-sanity-check.service \
+"
+
+FILES:${PN} = " \
+    ${bindir} \
+    ${systemd_system_unitdir} \
+"
+
+SYSTEMD_SERVICE:${PN} = " \
+    post-update-sanity-check.service \
+"
+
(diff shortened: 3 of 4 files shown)


recipes-core/rcu-sanity-check/files/post-update-sanity-check
```diff
@@ -0,0 +1,16 @@
...
+RCU_SERVICE_STATUS=$(systemctl is-active rcu-service)
+ETH0_STATUS=$(cat /sys/class/net/eth0/operstate)
+
+if [ "$RCU_SERVICE_STATUS" = "active" ]; then
+    mender commit
+    touch /data/rcu-service/firmware/update-succeeded
+    systemctl restart rcu-service
+else
+    mender rollback
+    touch /data/rcu-service/firmware/update-failed
```
HatsyRei, line 15: A note to review whether we want the script to reboot RCU in this case.

recipes-core/rcu-sanity-check/files/post-update-sanity-check
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+rm /data/rcu-service/firmware/update-staged
+rm /etc/rcu-mender-image-firstboot
+
+RCU_SERVICE_STATUS=$(systemctl is-active rcu-service)
+ETH0_STATUS=$(cat /sys/class/net/eth0/operstate)
```
HatsyRei, line 7: This is here in case we want to also sanity check eth0 is up.

recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Service to run RCU sanity check
+After=network.target
+ConditionPathExists=/etc/rcu-mender-image-firstboot
```
shlooiNI, line 4: Why not check if both `/etc/rcu-mender-image-firstboot` and `/data/rcu-service/firmware/update-staged` exist before running the script? That would solve the concern about this service running during TEZI update.
Alternatively, if we want to run sanity check on first boot by default, you can check for the `/data/rcu-service/firmware/update-staged` in post-update-sanity-check script before calling mender commit/rollback.

recipes-core/rcu-sanity-check/files/post-update-sanity-check
shlooiNI: I'm in favor of rebooting RCU in case of update failure. Reason is because if the update fails, ethernet may not be enabled, and hence users won't know what happened to the rack, and will also be unable to trigger reboot remotely.

recipes-core/rcu-sanity-check/files/post-update-sanity-check.service
HatsyRei: Good idea! I think I can add another ConditionPathExists with the pipe operator. That should work as expected.
https://stackoverflow.com/a/37313765

recipes-core/rcu-sanity-check/files/post-update-sanity-check
kengsooncheah: If eth0 is not connected for any reason (e.g. customer use eth1 instead). Will this detect a false failure?

recipes-core/rcu-sanity-check/files/post-update-sanity-check
```diff
@@ -0,0 +1,16 @@
...
+
+if [ "$RCU_SERVICE_STATUS" = "active" ]; then
+    mender commit
+    touch /data/rcu-service/firmware/update-succeeded
+    systemctl restart rcu-service
```
kengsooncheah, line 12: Why a restart is required?

The other question: why not let rcu-service itself do the checking and commit? We can make the rcu-service to check for the flag after it reaches operational state and commit. That ensures rcu-service is in good shape to serve.

shlooiNI: I think the concern is that rcu-service may fail before reaching the checking part

HatsyRei: @kengsooncheah @shlooiNI This does remind me of one issue with this approach..

The original reason we wanted a script to do the sanity check instead of rcu-service was so that if rcu-service is dead, we could still proceed to sanity check other things like image version, available interfaces, etc. The separate script will also be able to flag the update as failed (since rcu-service cannot do so in that case). However, the current approach with the script causes some circular dependency- We want the script to run after rcu-service starts (so that it can detect rcu-service is active), but it also needs to set the 'update-succeeded' flag before rcu-service consumes the flag. The restart there was to make sure that rcu-service picks up the flag and reports RCU last update status correctly.

Having rcu-service perform the sanity check can eliminate these problems. I'll need to change some logic with how rcu-service interprets the 'update-staged' flag. Combining with the /etc/rcu-mender-image-firstboot, I think we can make rcu-service differentiate between having to do sanity check and flagging the last update as a failed update.

HatsyRei: In short, I think we can go with the approach of getting rcu-service to do the sanity check- We'll still need a recipe which installs the first-boot flag. Combined with the 'update-staged' flag, rcu-service will be able to do the following, depending on which flags are present when RCU (re)boots:

1) 'first-boot' flag present but 'update-staged' flag not present
   - TEZI installation, or mender update which did not go through rcu-service gRPC API.
   - No sanity check is done, RCU last update status is unknown.
   - rcu-service removes the 'first-boot' flag.

2) 'first-boot' flag present and 'update-staged' flag present
   - Mender update through gRPC API
   - rcu-service will perform sanity check and mender commit if everything goes fine
   - If sanity check fails, rcu-service tries to set 'update-failed' flag (and rollback)
   - rcu-service removes the 'first-boot' flag.

3) 'first-boot' flag not present and 'update-staged' present
   - Mender update through gRPC API occurred, but rcu-service failed to run after rebooting into new image, or new image failed to boot properly. This results in RCU booting back into old image.
   - rcu-service should flag 'update-failed'

HatsyRei: @kengsooncheah @shlooiNI If the above logic looks good, I'll close this PR and open another PR for a recipe which only installs the first-boot flag.

shlooiNI: When will rcu-service perform the sanity check? If it performs it too early, it may not detect potential issue. But then if we try to leave the sanity check to later, rcu-service may have crashed, and the sanity check will not be run.

HatsyRei: @shlooiNI My plan is we do sanity check after rcu-service enters operational state. If rcu-service crashes and the sanity check fails to run, we'd consider the sanity check to have failed.

shlooiNI: In the case where rcu-service crashes, how would mender rollback be called?

HatsyRei: @shlooiNI In that case, the user will need to reboot the RCU. Without mender commit being executed by rcu-service, we expect the RCU to boot back to the old partition. In this case, the bootloader handles the rollback for us. For us, calling `mender rollback` explicitly prevents `mender commit` from being called in future, before the RCU reboots to the old image.

shlooiNI: ok, so means that user will need to physically access and reboot the rack, since remote access may not be available. If this is acceptable for customers, then I am ok with this suggestion.

recipes-core/rcu-sanity-check/files/post-update-sanity-check
HatsyRei: @kengsooncheah I think it will- this is part of the reason I decided to not include the check for now. Probably more work needed to get this right.

recipes-core/rcu-sanity-check/files/post-update-sanity-check
kengsooncheah: @HatsyRei, am I right that there are 3 files/flags that are used for communicating the status/state of the firmware update? They are 
1. /etc/rcu-mender-image-firstboot
2. /data/rcu-service/firmware/update-succeeded
3. /data/rcu-service/firmware/update-staged
Any that I missed?

HatsyRei: @kengsooncheah One more- `/data/rcu-service/firmware/update-failed`.

For sanity check, we will be relying mainly on the existence of `/etc/rcu-mender-image-firstboot` and `/data/rcu-service/firmware/update-staged`.

kengsooncheah: @HatsyRei it looks like these flags are mutually exclusive (no meaning if 2 of them appear a the same time), I wonder if we can collapse them into a single flag? say rcu-update-status and the content could be (installing, staged, verifying, succeeded, failed). I tried to figure out the overall process and drawn it here https://lucid.app/lucidchart/57148ac9-a38b-46a8-8b2a-fb878542d7f7/edit?viewport_loc=-11%2C-23%2C2219%2C1100%2C0_0&invitationId=inv_0fa1682b-2256-4df2-a3e5-d6691af1c2b3 for me to better understand the flow. Will talk to you offline.

kengsooncheah: Since we are going to change the approach to track the firmware update stages and need to rewrite this PR. I'm going to stop reviewing it. One thing to keep in mind is to make the design simple so that it's easy to debug.

PR #122, opened 2023-08-25
Install public key and enable Mender artifact verification

HatsyRei:
mender-client.bbappend recipe changes to install public key used for verifying signed RCU Mender images. The public key is empty here, and the real public key shall be used to replace this empty file in pipeline builds. After this change, Mender will try to verify signed images prior to installation. This can be disabled by manually removing the `ArtifactVerifyKey` attribute in mender.conf during runtime.

### DIFF
--- a/recipes-core/mender/files/mender.conf
+++ b/recipes-core/mender/files/mender.conf
@@ -1,0 +1,3 @@
+{
+    "ArtifactVerifyKey" : "/etc/mender/ateccgen2-rsa-public.key"
+}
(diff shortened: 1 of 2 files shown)


kengsooncheah (approved):
Try to understand the consequence of this change.

It will enforce the signature verification AFTER an RCU is updated to this image; for RCU's that are still running older image, it will not prevent the update, is that correct?

HatsyRei:
@kengsooncheah Yep, that is right. I'll install this on my RCU setup to further verify that this is the behavior. 

Edit: Behavior verified.

shlooiNI (approved):

PR #127, opened 2023-10-06
Alter method for determining current active partition

HatsyRei:
### Description
During driver implementation of rack firmware update API, it was recently discovered that performing updates multiple times without rebooting will cause the update to fail and rollback to the existing image version. Initial inspection revealed that the files which were supposed to be copied into the new partition were not copied correctly. 

Eventually, we managed to narrow it down to the use of `mender_boot_part` bootloader variable for determining the current active partition. Since `mender_boot_part` flips between the A/B Mender partitions when the first update is performed, this causes issues for the subsequent update taking place prior to rebooting the RCU. This problem should be applicable to both Mender updates triggered from rcu-service, as well as Mender updates triggered via SSH/serial.

### Solution
This PR replaces the use of `mender_boot_part` in state scripts with information found using the `mount` command. The `mount` command lists the partitions which are currently mounted on the RCU, and is able to tell if mmcblk0p2 or mmcblkp3 is mounted as root, making it ideal for the purpose required by state scripts. The expected behavior after this PR is merged is that the new image built from these changes should allow consecutive Mender updates prior to any reboots, to just work.

### Testing
Testing is conducted on the script level to ensure that the scripts run without errors, but a full test will require the image be built first then used for testing. Verified that root partition listed under `mount` changes between mmcblk0p2 and mmcblk0p3 accordingly after mender install + reboot, but not mender install alone.

### DIFF
--- a/recipes-mender/rcu-state-scripts/files/retain-credentials
+++ b/recipes-mender/rcu-state-scripts/files/retain-credentials
@@ -5,13 +5,8 @@
 
 echo "$(mender show-artifact): Running $(basename "$0")" >&2
 
-# Check if fw_printenv command is available
-if [ ! -x /usr/bin/fw_printenv ]; then
-    exit 1
-fi
-
 # Check current rootfs partition
-current=$(/usr/bin/fw_printenv mender_boot_part | awk -F = '{ print $2 }')
+current=$(cat /proc/cmdline | awk -F' ' '{ gsub("root=/dev/mmcblk0p", ""); print $2 }')
 
 # Deduce target rootfs partition based on current rootfs partition number
 if [ $current = "2" ]; then
--- a/recipes-mender/rcu-state-scripts/files/retain-credentials
+++ b/recipes-mender/rcu-state-scripts/files/retain-credentials
@@ -6,12 +6,12 @@
 echo "$(mender show-artifact): Running $(basename "$0")" >&2
 
 # Check current rootfs partition
-current=$(cat /proc/cmdline | awk -F' ' '{ gsub("root=/dev/mmcblk0p", ""); print $2 }')
+current=$(mount | awk '$3 == "/" {print $1}')
 
 # Deduce target rootfs partition based on current rootfs partition number
-if [ $current = "2" ]; then
+if [ $current = "/dev/mmcblk0p2" ]; then
     newroot=/dev/mmcblk0p3
-elif [ $current = "3" ]; then
+elif [ $current = "/dev/mmcblk0p3" ]; then
     newroot=/dev/mmcblk0p2
 else
     echo "Unexpected current root: $current" >&2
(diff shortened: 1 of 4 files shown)


recipes-mender/rcu-state-scripts/files/retain-credentials
```diff
@@ -5,13 +5,8 @@
...
 echo "$(mender show-artifact): Running $(basename "$0")" >&2
 
-# Check if fw_printenv command is available
-if [ ! -x /usr/bin/fw_printenv ]; then
-    exit 1
-fi
-
 # Check current rootfs partition
-current=$(/usr/bin/fw_printenv mender_boot_part | awk -F = '{ print $2 }')
+current=$(cat /proc/cmdline | awk -F' ' '{ gsub("root=/dev/mmcblk0p", ""); print $2 }')
```
HatsyRei: A bit of info on the awk command. The `-F` specifies the separator, which in this case is a whitespace. `gsub` is then used to delete the `root=/dev/mmcblk0p` string, and $2 is the second group after the separator is applied to /proc/cmdline.

This is an example of the raw output of /proc/cmdline:
```
pci=nomsi root=/dev/mmcblk0p3 ro rootwait
```

so we grab `root=/dev/mmcblk0p3`, remove `root=/dev/mmcblk0p` and is left with `3` 

Please let me know if there's a better way... :D

shlooiNI: is /proc/cmdline available by default on rcu image? just want to confirm that it is something that is already available.

shlooiNI (approved):

recipes-mender/rcu-state-scripts/files/retain-credentials
HatsyRei: @shlooiNI I believe proc is a default part of the Linux kernel, so it should be readily available. Seems like it goes a long way back too...
https://elixir.bootlin.com/linux/latest/source/Documentation/filesystems/proc.rst

kengsooncheah: Not sure how frequent the command line is changed. If you want a more consistent output, maybe can try `mount`, e.g. `mount | awk '$3 == "/" {sub(/\/dev\/mmcblk0p/,"", $1); print $1}'` or you can even skip the substitution,

```
current=$(mount | awk '$3 == "/" {print $1}'

# Deduce target rootfs partition based on current rootfs partition number
if [ $current = "/dev/mmcblk0p2" ]; then
    newroot=/dev/mmcblk0p3
...
```

kengsooncheah (approved):

recipes-mender/rcu-state-scripts/files/retain-credentials
HatsyRei: @kengsooncheah I like the no sub solution! Updated

PR #128, opened 2023-10-09
Fix RCU Image default password hash

HatsyRei:
Issue found during testing of new RCU. Failed to login to image flashed via TEZI using default password. After investigating the issue, turns out the `$` character needs to be escaped. Tested the usermod command on Gamora and verified that password is correct after the change.

### DIFF
--- a/recipes-images/images/ateccgen2-base-image.inc
+++ b/recipes-images/images/ateccgen2-base-image.inc
@@ -14,7 +14,7 @@ COPY_LIC_DIRS ?= "1"
 
 # Configure base image root account default password
 inherit extrausers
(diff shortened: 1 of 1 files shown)


kengsooncheah:
How would this issue affect the RCU that has been built? Do we need to give heads up to test engineers?

kengsooncheah (approved):

HatsyRei:
@kengsooncheah Looked into this with @shlooiNI earlier. Seems like this problem won't affect already built racks as they have been using 23.0.1 releases. The hash was introduced during Dunfell -> Kirkstone migration, which happened after 23.0.1.

shlooiNI (approved):

HatsyRei:
Anyway @CHrislian-NI, this is the root cause of why we weren't able to login to the image in failed update scenario. Default pwd was messed up.

PR #129, opened 2023-10-27
Remove Wifi, BT, sound and other unused components from image

HatsyRei:
PR to remove unused components from RCU image. ~10MB reduction in terms of image size. 

Testing: Built and tested image on RCU + fanout board setup. Verified that image no longer contains BT and Wifi related kernel modules and packages.

### DIFF
--- a/conf/machine/include/apalis-imx8.inc
+++ b/conf/machine/include/apalis-imx8.inc
@@ -1,0 +1,3 @@
+DISTRO_FEATURES:remove = " 3g alsa bluetooth wifi nfc "
+MACHINE_FEATURES:remove = " alsa bluetooth wifi "
+
--- a/recipes-kernel/linux/files/ateccgen2.cfg
+++ b/recipes-kernel/linux/files/ateccgen2.cfg
@@ -18,7 +18,4 @@ CONFIG_CIFS=m
 # CONFIG_BT is not set
 # CONFIG_RFKILL is not set
 # CONFIG_CFG80211 is not set
-# CONFIG_SOUND is not set
 # CONFIG_CAN is not set
-# CONFIG_WIRELESS is not set
-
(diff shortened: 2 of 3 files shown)


conf/machine/include/apalis-imx8.inc
```diff
@@ -0,0 +1,3 @@
+DISTRO_FEATURES:remove = " 3g alsa bluetooth wifi nfc "
+MACHINE_FEATURES:remove = " alsa bluetooth wifi "
```
HatsyRei, line 2: For future reference. Distro and machine features cannot be removed in image recipe. 
https://yocto.yoctoproject.narkive.com/zQ11Cl29/best-way-to-remove-distro-features

This entry under local.conf allows including additional conf data:
https://github.com/ni/meta-ateccgen2/blob/main/buildconf/local.conf#L275

recipes-kernel/linux/files/ateccgen2.cfg
```diff
@@ -14,3 +14,8 @@ CONFIG_BMP280=y
 CONFIG_BMP280_I2C=y
 CONFIG_BMP280_SPI=y
 CONFIG_CIFS=m
+# CONFIG_WLAN is not set
+# CONFIG_BT is not set
+# CONFIG_RFKILL is not set
+# CONFIG_CFG80211 is not set
+# CONFIG_CAN is not set
```
HatsyRei, line 21: For some reason, removing CONFIG_SOUND from defconfig will cause some issues when compiling gpio-switch.c, in a similar fashion as in:
https://www.mail-archive.com/linux-kernel@vger.kernel.org/msg2541557.html

Temporarily removed omission of CONFIG_SOUND, but note that RCU image does not require it.

shlooiNI (approved):

kengsooncheah (approved):

PR #139, opened 2024-07-23
Prevent emc2303 fan driver from reporting 1875 RPM reading

HatsyRei:
Spent some time reviewing the [emc2303 datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MSLD/ProductDocuments/DataSheets/EMC2301-2-3-5-Data-Sheet-DS20006532A.pdf) as well as the existing patch which adds driver support for emc2303 in our meta layer.

EMC2303 registers involved in reading fan tach are 0x3e (high byte) and 0x3f (low byte). 
image

Default value for these two bytes are 0xff and 0xf8 respectively (i.e., 1111 1111 and 11111000). The reason for this is because 8 bits in the high byte and only 5 bits in the low byte are involved in determining fan tach reading:
image

The driver combines high byte and low byte via the following operation:
```cpp
*output = ((u16)high_byte << 5) | (lo_byte >> 3);
```
where high byte is leftshifted by 5 to make way for the 5 meaningful bits in low byte, then combined with bitwise OR.

Taking into account the default values for these two bytes, this would make the default tach reading 0001 1111 1111 1111 (0x1fff). It is also the highest possible register reading which can possibly be read from these two registers. 

The way these register readings are used to calculate the actual RPM value is listed here:
image

Using the default register values, the watered-down version of the driver operation is show in code below:
```cpp
#define FAN_TACH_EDGES 5
#define FAN_POLES      2
#define FAN_MULTIPLIER 4
#define FAN_CLK_FREQ   32000

/* Apply Equation 4-2 from datasheet */
#define FAN_RPM_FACTOR (((FAN_TACH_EDGES - 1) * FAN_MULTIPLIER * FAN_CLK_FREQ * 60) / FAN_POLES) // 15360000

uint16_t tach = ((uint16_t)0xff << 5) | (0xf8 >> 3); // This gives 0x1fff
int RPM = FAN_RPM_FACTOR / tach; // 1875
```
This is one of the ways how the value of 1875 can be obtained. It also means that 1875 is the lowest possible RPM reading which can be obtained from the driver (given the existing fan configuration). When modifying the kernel driver to print the actual tach register values when fan is not powered, the value of 8188 (or 0x1ffc) is obtained instead of 0x1fff.
image
The exact reason for this is still unknown, but 0x1ffc is the smallest denominator which can result in RPM reading of 1875. 0x1ffb, for example, results in 1876. 

This PR proposes to allow the driver to report a RPM reading of 0 when fan tach is found to be >= 0x1ffc. (Reason being values 0x1ffc to 0x1fff will result in the 1875 reading)

##Testing

Testing conducted on Widow:
image

### DIFF
--- a/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
+++ b/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
@@ -494,7 +494,7 @@ index 00000000..b2ae311
 +	if (status < 0)
 +		return status;
 +
-+	if (fan->tach != 0)
++	if ((fan->tach != 0) && (fan->tach != 0x1fff))
 +		rpm = FAN_RPM_FACTOR / fan->tach;
 +	return sprintf(buf, "%d\n", rpm);
 +}
--- a/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
+++ b/recipes-kernel/linux/files/0001-hwmon-add-support-for-SMSC-EMC2303.patch
@@ -494,7 +494,7 @@ index 00000000..b2ae311
 +	if (status < 0)
 +		return status;
 +
-+	if ((fan->tach != 0) && ((fan->tach & 0x1fff) != 0x1fff))
++	if ((fan->tach != 0) && ((fan->tach & 0x1ffc) != 0x1ffc))
 +		rpm = FAN_RPM_FACTOR / fan->tach;
 +	return sprintf(buf, "%d\n", rpm);
 +}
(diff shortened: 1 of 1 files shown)


shlooiNI (commented):
I see that both `REG_FAN_STATUS = 0x24` and `REG_FAN_STALL_STATUS = 0x25` are defined in the global registers enum, but are unused. According to the data sheet, depending on the fan mode, we should be able to read these registers to determine if a fan has stalled. But I'm not sure what mode we are using.
image
image

HatsyRei:
@shlooiNI Managed to try out the 25h register. 
image

I believe it's not working because our fans are in Fan Speed Control mode instead of Direct Setting mode. We'll have to rely on the tach register readings.

shlooiNI (approved):

PR #142, opened 2024-10-07
Add device tree overlays to support multirack IO expanders

HatsyRei:
PR to address [US 2896515](https://dev.azure.com/ni/DevCentral/_workitems/edit/2896515).

I/O expander locations currently set by HW at I2C4, 24h, 25h and 21h for rack 1, 2 and 3 respectively. 

Testing was done on prototype connected to PEP I2C, so we'll need to test this again once the real multirack setup is available.


shlooiNI (approved):
Not an expert, but nothing looks out of place to me.

shlooiNI (approved):

